PR #16, opened 2020-12-28
update to use the @lightningjs namespace

chrisdeely:


chiefcll:
Will open a new one to do this

Issue #18, opened 2021-01-30
Feature request: Infinite scrolling style of wrapping


edatencio-velope:
Hello!

Currently, the wrapping style of the row component is scrolling back to the first item. It would be very nice if you could add a wrapping style that makes the row appear infinite. It doesn't need to duplicate items because we are using enough to fill the screen width.

chiefcll:
We almost have this PR ready... Hopefully early next week

vmend:
@chiefcll Hello, any news on this? Thank you!

PR #32, opened 2021-05-17
feat: Carouse Row

chelseasimek:
To address issue #18 

TODO: 

- fix scrollIndex which does not work
- fix selected index nav left
- update left navigation behavior


erautenberg:
Closing as this is outdated. This can be re-visited after the new Focus Manager and Row components are added to the repo.

PR #45, opened 2021-09-10
feat(Base): add Base component

THoj13:
- add `Base` component

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,20 +1,20 @@
 /**
-* Copyright 2020 Comcast Cable Communications Management, LLC
-*
-* Licensed under the Apache License, Version 2.0 (the "License");
-* you may not use this file except in compliance with the License.
-* You may obtain a copy of the License at
-*
-* http://www.apache.org/licenses/LICENSE-2.0
-*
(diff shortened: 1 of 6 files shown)


index.js
```diff
@@ -23,3 +23,4 @@ export { default as FocusManager } from './components/FocusManager/index.js';
 export { default as Icon } from './components/Icon/index.js';
 export { default as MarqueeText } from './components/MarqueeText/index.js';
 export { default as Row } from './components/Row/index.js';
+export { default as Base } from './components/Base';
```
erautenberg, line 26: I'd like to mimic the inner source repo here and create folders for elements vs layout vs mixins, then have each of those folders with their own index.js to handle the exports in that folder.

erautenberg (changes requested):

index.js
anthony9187: is it cool if we tackle splitting those out in a separate PR? I'm working on it in my layout updates

anthony9187 (approved):

erautenberg (approved):

PR #46, opened 2021-09-10
feat(mixins): add withTags

THoj13:
- add withTags mixin

### DIFF
--- a/mixins/withTags/index.js
+++ b/mixins/withTags/index.js
@@ -15,6 +15,7 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  */
+
 function getPropertyDescriptor(path) {
   return {
     get() {
--- a/mixins/withTags/index.js
+++ b/mixins/withTags/index.js
@@ -1,0 +1,52 @@
+/**
+ * Copyright 2020 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


mixins/withTags/index.js
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+function getPropertyDescriptor(path) {
```
erautenberg, line 19: super nit-picky, but can we add a newline here, and maybe add a rule to the linter to have a blank line between comment blocks like this?

erautenberg (approved):

mixins/withTags/index.js
THoj13: I've added the new line. I'd like to add the linting for this in a separate PR, as I think there's additional setup needed first. Currently, `eslint` is not installed as a dependency so I don't believe `.eslintrc` is being used. The current `.eslintrc` also has rules for React/jsx, which I assume won't be required in this repo.

anthony9187: that's correct. i have a branch with the updates to eslint in it that i can hopefully get into a PR soon. or we can split it out to it's own PR

anthony9187 (approved):

PR #59, opened 2021-09-20
feat(mixins): add withLayout mixin

THoj13:
Adds `withLayout` mixins.

To better display the effects of this mixin, the storybook should be updated to use `withMetaData` and `MetadataTile` once those have been added to this repo.

PR #61, opened 2021-09-21
feat(Base): wrap Base component with mixins

THoj13:
Wraps the `Base` component with all the mixins. 

~**Note:** This PR is comparing against `feat/with-layout` as that is what `feat/base-mixins` was based off of. It should be merged after https://github.com/rdkcentral/Lightning-UI-Components/pull/59 has been merged.~


elements/Base/Base.test.js
```diff
@@ -22,13 +22,12 @@ import Base from '.';
 const createComponent = TestUtils.makeCreateComponent(Base);
 
 describe('Base', () => {
-  let component, testRenderer;
+  let testRenderer;
 
   beforeEach(() => {
-    [component, testRenderer] = createComponent();
+    [, testRenderer] = createComponent();
```
anthony9187, line 28: can we remove the comma here?

THoj13: If the comma is removed, `testRender` would be assigned the first returned value from `createComponent()` (what was previously `component`). The comma allows assigning the second value without needing to declare a variable for the first value. `component` wasn't being referenced anywhere in the test, so I removed it for linting. This syntax is a bit odd looking though, so I could also see just leaving it as is and have the linter ignore that line.

anthony9187 (approved):

chelseasimek (approved):
Looks good!

PR #64, opened 2021-09-21
Feat(mixins): add withAnnouncer

THoj13:
- adds `withAnnouncer` mixin
- updates storybook preview to wrap components with `withAnnouncer`


THoj13:
@erautenberg Should `Announcer` be removed with this update or is there plans to support it along with `withAnnouncer`?

chelseasimek (changes requested):
This same code is in elements right now as well...we should deprecate that in this PR as well

THoj13:
> This same code is in elements right now as well...we should deprecate that in this PR as well

@chelseasimek Sounds good. `elements/Announcer` has now been removed in favor of `mixins/withAnnouncer`

chelseasimek (approved):

anthony9187 (approved):

PR #66, opened 2021-09-22
feat: enable eslint

anthony9187:
This PR updates the internal `lint` scripts to run eslint instead of prettier. It also fixes all of the existing linting errors in the components.

### DIFF
--- a/elements/Keyboard/Keyboard.test.js
+++ b/elements/Keyboard/Keyboard.test.js
@@ -19,6 +19,7 @@
 import TestUtils from '../../test/lightning-test-utils';
 import Keyboard, { KEYBOARD_FORMATS } from '.';
 import Key from './Key';
+// import { getHexColor } from '../../Styles';
 
 const icon = TestUtils.pathToDataURI('assets/images/ic_lightning_white_32.png');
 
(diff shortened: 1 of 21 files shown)


elements/Keyboard/Keyboard.test.js
```diff
@@ -93,16 +94,19 @@ describe('Key', () => {
...
     key._focus();
 
     setTimeout(() => {
+      // are these test values correct?
       expect(key.color).toBe(getHexColor('ECECF2'));
       expect(key.scale).toBe(1.12);
       expect(key._Title.color).toBe(4060086272);
-      done();
-    });
+    }, 0);
```
anthony9187: @chelseasimek can you take a look at this when you've got some free time? I'm not sure where these expected values came from, including the scale(as far as I know, the key shouldn't scale on focus?). 

The `_focus` method is part of `Button` and not `Key`, so I'm also wondering if coverage in this test suite is even necessary

elements/Announcer/Speech.test.js
```diff
@@ -221,7 +222,7 @@ describe('Speech', () => {
     speakError = 'network';
     const warnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {});
     jest.useFakeTimers();
-    const speaking = Speech(['Hello', 'World']);
+    const speaking = Speech(['Hello', 'World']); // eslint-disable-line no-unused-vars
```
THoj13: I think this also would work, and follow lint rules, by invoking `Speech` and not assigning the return to a variable.
```js
Speech(['Hello', 'World']);
```

THoj13 (approved):

elements/Announcer/Speech.test.js
anthony9187: I'll give that a try!

erautenberg (approved):

chelseasimek (approved):

PR #67, opened 2021-09-22
feat(Base): add default theme constant

THoj13:
- composed `DEFAULT_THEME` with existing properties defined in `Styles` dir so that component's style functions have a theme to reference
```js
// SomeComponent.styles.js
export default theme => ({
  shadow: theme.materials.glow,
})
```
- added `Styles` dir to list of paths to directories that Jest should use to search for files in.  As of this change, `Styles` still will be excluded from coverage, but this allows testing any functions in `Styles/`. This was added to allow testing the `createTheme` function in isolation.
- split constants used to compose `DEFAULT_THEME` to separate files
- restructured some files in Styles and mixins directories to avoid circular dependency with `Styles` export and `createTheme` function
    - Issue: `Styles/Styles.js` was importing an re-exporting values from `createTheme.js` `createTheme.js` was importing `CORNER_RADIUS` from `Styles.js`
    - moved `CORNER_RADIUS` constant to be exported from `Styles/Materials.js`
    - moved `createTheme` to `mixins` dir. As of now,  exports from `createTheme.js` are only used within the mixins dir (`withStyles` and `withTheme`)
    - deleted `Styles/Styles.js` which just imported values from all the files within the Styles dir and re-exported them. `Styles/index.js` then exported all values from `Styles.js`. This change instead  has `Styles/index.js` handle importing all values from within Styles dir and re-exporting them.
    - Result: nothing in the `Styles` dir requires values from `createTheme.js`, `createTheme` still imports values from `Styles`


chelseasimek (approved):

Styles/Materials.js
```diff
@@ -17,10 +17,39 @@
...
+    inactive: 40
+  }
+};
+
+export const CORNER_RADIUS = {
+  large: 24,
+  medium: 16,
+  small: 8,
+  xsmall: 4
+};
```
chelseasimek: Alpha and corner radius don't belong in Materials.js - I would take a look at the themeing work that Josh is working on and try to set up the Styles folder similar to that so it's easier to roll over to that work

anthony9187 (approved):
looks good!

THoj13:
Waiting to merge until theme work has been completed in inner source repo.

THoj13:
Closing this for now since a lot of this will likely not be applicable once the theming work is in place.

PR #79, opened 2021-10-12
feat(Slider): add Slider element

### LINKED ISSUE #75: Add Slider element
Add a Slider - a progress control component, similar to the DOM's `input[type="range"]`

THoj13:
- moved `FadeShader` to `textures/` , exported from root
- added `Slider` element


textures/index.js
```diff
@@ -0,0 +1,200 @@
...
+      ctx.closePath();
+      ctx.stroke();
+      ctx.fill();
+
+      cb(null, { source: canvas, w, h, direction });
+    };
+  }
+}
+
+export { FadeShader, Circle, Arrow };
```
anthony9187: can we split these out to separate files and export them from the textures/index.js?

THoj13: Can do!

anthony9187 (approved):

erautenberg (changes requested):
This is on a white background for me. Is that true for anyone else? Shouldn't it be a white slider on black background?

anthony9187:
> This is on a white background for me. Is that true for anyone else? Shouldn't it be a white slider on black background?

I'm seeing this as well, can we update this so the story matches the others?

THoj13:
> > This is on a white background for me. Is that true for anyone else? Shouldn't it be a white slider on black background?
> 
> I'm seeing this as well, can we update this so the story matches the others?

Updated to use dark background

erautenberg (approved):

PR #81, opened 2021-10-14
feat(Column): update with latest bug fixes

THoj13:
update Column with latest bug fixes from LUI


anthony9187 (approved):

erautenberg:
Posted this on the other PR:
Looks like this was last updated in the inner source repo 5 days ago, but this is from 7 days ago. Just double check you got all of the changes?

Might just be documentation changes

layout/Column/Column.stories.js
```diff
@@ -174,7 +174,7 @@ export const Plinko = args =>
             {
               type: Row,
               h: 40,
-              w: 900,
+              w: 1920 - 180, // x offset from preview.js * 2,
```
erautenberg: I _think_ the w is actually 1920 - 160 now and the stage should have been updated to an x position of 80 instead of 90 (might need to double check Row and Announcer too): https://github.comcast.com/Lightning/lightning-ui/pull/710

THoj13: Sounds good, updated the widths to match LUI.

erautenberg (approved):

PR #86, opened 2021-10-27
fix(package): add Styles textures and mixins to package

THoj13:
- adds exported modules from `Styles`,  `textures`,  and `mixins` to package

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,11 +15,11 @@
     "index.js",
     "Styles.js",
     "{bin,elements,layout,mixins,Styles,textures,utils}/**/*",
-    "!{elements,layout,mixins,Styles}/**/*.stories.js",
-    "!{elements,layout,mixins,textures,utils}/**/*.test.js",
+    "!{elements,layout,mixins,Styles,textures,utils}/**/*.stories.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -14,12 +14,12 @@
...
-    "!{elements,layout,utils}/**/*.test.js",
-    "!elements|layout/lightning-test-renderer.js",
-    "!elements|layout/lightning-test-utils.js",
-    "!elements|layout/**/__snapshots__",
+    "{bin,elements,layout,mixins,Styles,textures,utils}/**/*",
+    "!{elements,layout,mixins,Styles}/**/*.stories.js",
+    "!{elements,layout,mixins,textures,utils}/**/*.test.js",
+    "test/lightning-test-renderer.js",
+    "test/lightning-test-utils.js",
+    "!{elements,layout,textures}/**/__snapshots__",
```
anthony9187: i understand that not all of these directories contain stories, snapshots, etc. but I think it might be a good idea to use the same globs for lines 18, 19, and 22 in case these files get added in the future, and just for consistency

anthony9187 (approved):
🎉

erautenberg (approved):

PR #97, opened 2022-07-12
feat: update FocusManager Row Column and dependencies to inner source v4.47.2

THoj13:
Update FocusManager, Row, Column, and all dependencies of those components (any file that any of those components import) to the latest available from the inner source repo (v.4.47.2)


THoj13:
Review https://github.com/rdkcentral/Lightning-UI-Components/pull/98 instead of this one.

PR #113, opened 2023-03-01
feat: add demo app

joshhowenstine:
Adds basic demo application with browse page with mocked data from [TMDB](https://www.themoviedb.org/?language=en-US).

**Testing notes:**
1. `yarn install`
2. `yarn workspace demo start`

PR #114, opened 2023-03-01
fix: add Metrological line to license info in NOTICE

erautenberg:
@mhughesacn I saw your comment [here](https://github.com/rdkcentral/Lightning-UI-Components/pull/113#issuecomment-1450618792)
> Also please append credit to NOTICE for the Metrological use:
>
> Copyright 2020 Metrological
> Licensed under the Apache License, Version 2.0

Is this all that's needed? Thank you!

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,6 +1,8 @@
 [Lightning-UI-Components]
 Copyright 2020 Comcast Cable Communications Management, LLC
 
+Copyright 2020 Metrological
+
 Licensed under the Apache License, Version 2.0 (the "License");you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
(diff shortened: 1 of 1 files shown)


mhughesacn:
Hi @erautenberg
Not exactly: PR#113 has been changed by Josh, and points 1, 2 have been fixed.  Point 4 not relevant as file deleted, and point 5 (NOTICE) is also no longer relevant because the copyright was changed to Comcast, which is why BlackDuck stopped complaining.  (If it is a Metrological file, then it would be better to leave the M.copyright, and add the line to NOTICE.  Also, if the addition to NOTICE is needed, then it should go at the very end of the file as that's how we do it.)
Point 3 is still open: to be certain that all new icons are original to Comcast (or Metrological, Sky).  Otherwise, I need the source and credit is usually needed.  I just ran a special scan and BlackDuck doesn't find any matches to the icons in PR#113 but I would like confirmation.
Many thanks,
Martin

erautenberg:
@mhughesacn thanks so much for the quick reply! I totally understand, I will close the PR then.

Also, @joshhowenstine and I talked through his PR yesterday and we are actually going to swap the Sky theme for a new, totally original one, so all of those assets will be completely removed. All other assets in our repo either are pulled directly via URL from TMDB or are original Comcast assets. We also have a credit line in our docs for TMDB.

PR #121, opened 2023-03-02
test(TextBox): restore skipped test

soumyaloka:
Restore skipped test in Textbox and add unit test to increase the coverage.

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/TextBox/TextBox.test.js
+++ b/packages/@lightningjs/ui-components/src/components/TextBox/TextBox.test.js
@@ -173,20 +173,47 @@ describe('TextBox', () => {
         name: 'Test',
         typography: {
           body1: {
-            fontFace: 'TestFont2',
-            fontSize: 15,
-            fontWeight: 250,
-            letterSpacing: 0,
-            lineHeight: 50,
-            verticalAlign: 'middle'
+            fontFace: 'TestFont2'
           }
         }
+      };
+      const baseThemeValues = {
+        typography: {
+          body1: {
+            fontSize: 25,
+            fontWeight: 300,
+            lineHeight: 40,
+            verticalAlign: 'middle',
+            textBaseline: 'bottom'
+          }
+        }
+      };
+      const lightningDefaultValues = {
+        fontSize: 40,
+        fontStyle: 'normal',
+        textAlign: 'left',
+        verticalAlign: 'top',
+        wordWrap: true,
+        maxLines: 0,
+        maxLinesSuffix: '..',
+        wordWrapWidth: 0
       };
       context.setTheme(testTheme);
       textBox.content = 'Hello Text Defaults';
       await textBox.__updateSpyPromise;
       expect(context.theme.name).toBe('Test');
-      expect(textBox.style.textStyle).toMatchObject(testTheme.typography.body1);
+      expect(textBox._Text.text.fontFace).toBe(
+        testTheme.typography.body1.fontFace
+      );
+      expect(textBox._Text.text.fontSize).toBe(
+        baseThemeValues.typography.body1.fontSize
+      );
+      expect(textBox._Text.text.maxLinesSuffix).toBe(
+        lightningDefaultValues.maxLinesSuffix
+      );
+      expect(textBox._Text.text.fontStyle).toBe(
+        lightningDefaultValues.fontStyle
+      );
     });
   });
 
--- a/packages/@lightningjs/ui-components/src/components/TextBox/TextBox.test.js
+++ b/packages/@lightningjs/ui-components/src/components/TextBox/TextBox.test.js
@@ -195,6 +195,7 @@ describe('TextBox', () => {
       textBox.content = 'Hello Text Defaults';
       await textBox.__updateSpyPromise;
       expect(context.theme.name).toBe('Test');
+      expect(textBox.style.textStyle).toMatchObject(testTheme.typography.body1);
     });
   });
 
--- a/packages/@lightningjs/ui-components/src/components/TextBox/TextBox.test.js
+++ b/packages/@lightningjs/ui-components/src/components/TextBox/TextBox.test.js
@@ -169,22 +169,12 @@ describe('TextBox', () => {
     });
 
     it('should get TextDefaults when switching themes', async () => {
+      const baseTheme = context.theme.typography.body1;
       const testTheme = {
         name: 'Test',
         typography: {
           body1: {
             fontFace: 'TestFont2'
-          }
-        }
-      };
-      const baseThemeValues = {
-        typography: {
-          body1: {
-            fontSize: 25,
-            fontWeight: 300,
-            lineHeight: 40,
-            verticalAlign: 'middle',
-            textBaseline: 'bottom'
           }
         }
       };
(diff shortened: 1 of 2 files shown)


packages/@lightningjs/ui-components/src/components/TextBox/TextBox.test.js
```diff
@@ -168,14 +168,37 @@ describe('TextBox', () => {
...
+          }
+        }
+      };
+      context.setTheme(testTheme);
+      textBox.base = context.theme;
+      textBox.content = 'Hello Text Defaults';
+      await textBox.__updateSpyPromise;
+      expect(context.theme.name).toBe('Test');
+      expect(textBox.style.textStyle).toEqual({
+        ...context.theme.typography.body1
```
erautenberg: I'm not sure if this is the right test case. Don't we want to confirm that textBox.style.textStyle matches testTheme.body1 to check that the theme was updated properly?

packages/@lightningjs/ui-components/src/components/TextBox/TextBox.test.js
```diff
@@ -168,14 +168,35 @@ describe('TextBox', () => {
...
+            fontSize: 15,
+            fontWeight: 250,
+            letterSpacing: 0,
+            lineHeight: 50,
+            verticalAlign: 'middle'
+          }
+        }
+      };
+      context.setTheme(testTheme);
+      textBox.base = context.theme;
```
ImCoolNowRight: Line 194 isn't doing anything and can be removed.

packages/@lightningjs/ui-components/src/components/TextBox/TextBox.test.js
```diff
@@ -168,14 +168,34 @@ describe('TextBox', () => {
...
+        name: 'Test',
+        typography: {
+          display1: {
+            fontFace: 'TestFont1',
+            fontSize: 35,
+            fontWeight: 500,
+            letterSpacing: 1,
+            lineHeight: 100,
+            verticalAlign: 'middle'
+          },
```
ImCoolNowRight: Do we need this if we aren't checking for any of these values?

soumyaloka: expect(textBox.style.textStyle).toMatchObject(testTheme.typography.body1);
In the above test it make sure to match the object that includes everything defined in testTheme.typography.body1.
Or do you think we don't need to test this?

ImCoolNowRight: No, I think that's fine! I'm referring to the `display1` object. There's no test for that, right?

soumyaloka: Oh, yeah. I misread it as body1. Yeah I can remove that.

ImCoolNowRight (approved):

packages/@lightningjs/ui-components/src/components/TextBox/TextBox.test.js
```diff
@@ -168,14 +168,26 @@ describe('TextBox', () => {
...
 
-    // TODO: Fix reference to Base
-    // it('should get TextDefaults when switching themes', async () => {
-    //   context.setTheme(Base);
-    //   textBox.base = context.theme;
-    //   textBox.content = 'Hello Text Defaults';
-    //   await textBox.__updateSpyPromise;
-    //   expect(textBox._Text.text.maxLinesSuffix).toBe('..');
-    // });
+    it('should get TextDefaults when switching themes', async () => {
```
jazzyw, line 171: By TextDefaults does it mean the [Lightning default text styles](https://lightningjs.io/docs/#/lightning-core-reference/RenderEngine/Textures/Text?id=text)? If so, would this test be testing if on theme change, text props like `fontSize` is the default size of 40 rather than updating that text style property?

jazzyw: I was asking/thinking something perhaps like this:

```js
    it('should get TextDefaults when switching themes', async () => {
      const testTheme = {
        name: 'Test',
        typography: {
          body1: {
            fontFace: 'TestFont2'
          }
        }
      };
      context.setTheme(testTheme);
      textBox.content = 'Hello Text Defaults';
      await textBox.__updateSpyPromise;
      expect(context.theme.name).toBe('Test');


      /* new addition */
      const baseThemeValues = {
        typography: {
          body1: {
            fontSize: 25,
            fontWeight: 300,
            lineHeight: 40,
            verticalAlign: 'middle',
            textBaseline: 'bottom'
          }
        }
      }

      const lightningDefaultValues = {
        fontSize: 40,
        fontStyle: 'normal',
        textAlign: 'left',
        verticalAlign: 'top',
        wordWrap: true,
        maxLines: 0,
        maxLinesSuffix: '..',
        wordWrapWidth: 0
      }
      
      expect(textBox._Text.text.fontFace).toBe(testTheme.typography.body1.fontFace)
      // since we only updated the fontFace, the text props defined in the Base theme should apply to the new theme
      expect(textBox._Text.text.fontSize).toBe(baseThemeValues.typography.body1.fontSize)
      // those text props not defined in the Base theme, should use the Lightning defaults
      expect(textBox._Text.text.maxLinesSuffix).toBe(lightningDefaultValues.maxLinesSuffix);
      expect(textBox._Text.text.maxLines).toBe(lightningDefaultValues.maxLines);
    });
```

soumyaloka: Got it, Nice. I can add that.

packages/@lightningjs/ui-components/src/components/TextBox/TextBox.test.js
```diff
@@ -168,14 +168,53 @@ describe('TextBox', () => {
...
+      const testTheme = {
+        name: 'Test',
+        typography: {
+          body1: {
+            fontFace: 'TestFont2'
+          }
+        }
+      };
+      const baseThemeValues = {
```
jazzyw: Thanks for updating! And sorry I know I suggested this but do you think it would be better to reference the Base/default theme rather than create the baseTheme object ourselves? That way if the default theme is updated then we don't need to update the object in this unit test.

Something like this:

```js
   it('should get TextDefaults when switching themes', async () => {
      // storing default theme (since the theme hasn't been switched yet)
      const baseTheme = context.theme.typography.body1;

      const testTheme = {
        name: 'Test',
        typography: {
          body1: {
            fontFace: 'TestFont2'
          }
        }
      };

      const lightningTextDefault = {
        fontSize: 40,
        fontStyle: 'normal',
        textAlign: 'left',
        verticalAlign: 'top',
        wordWrap: true,
        maxLines: 0,
        maxLinesSuffix: '..',
        wordWrapWidth: 0
      };

      context.setTheme(testTheme);
      textBox.content = 'Hello Text Defaults';
      await textBox.__updateSpyPromise;

      expect(context.theme.name).toBe('Test');
      expect(textBox._Text.text.fontFace).toBe(
        testTheme.typography.body1.fontFace
      );
      expect(textBox._Text.text.fontSize).toBe(baseTheme.fontSize);
      expect(textBox._Text.text.maxLinesSuffix).toBe(
        lightningTextDefault.maxLinesSuffix
      );
      expect(textBox._Text.text.fontStyle).toBe(
        lightningTextDefault.fontStyle
      );
    });
```

jazzyw: On that note, can we reference the Lightning default text props without defining the object ourselves?

soumyaloka: Yeah, we can do that for baseTheme, but not sure if we can test without referencing object for LightningDefaultValues?

jazzyw: Sure then let's change it for the baseTheme but keep the lightningDefault object

jazzyw: And instead of `baseTheme` you can also do `defaultTheme` whichever makes the most sense

jazzyw (approved):

PR #127, opened 2023-03-06
fix(Tooltip): fixed args url issue by adding sharedargactions

soumyaloka:
## Description

In this PR, I have updated add argActions to Basic tooltip and Long title story.

## Testing

1. Navigate to the 'Tooltip Basic' component
2. Go to unfocused mode and set the delayVisible in milliseconds like 2000
3. Open the story only in a new tab (hit the arrow button in the corner)
4. Change the mode to focused and update delayVisible then observe the component if the title visible is delayed for the specified seconds, repeat the same with timeVisible.

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Tooltip/Tooltip.stories.js
+++ b/packages/@lightningjs/ui-components/src/components/Tooltip/Tooltip.stories.js
@@ -58,23 +58,40 @@ export default {
   }
 };
 
-export const Basic = () =>
+const sharedArgActions = {
+  argActions: {
+    delayVisible: (delayVisible, component) => {
+      component.tag('Tooltip').delayVisible = delayVisible;
+    },
+    timeVisible: (timeVisible, component) => {
+      component.tag('Tooltip').timeVisible = timeVisible;
+    }
+  }
+};
+
+export const Basic = args =>
   class Basic extends lng.Component {
     static _template() {
       return {
         Tooltip: {
           type: TooltipComponent,
-          x: 1280 / 2
+          x: 1280 / 2,
+          delayVisible: args.delayVisible,
+          timeVisible: args.timeVisible
         }
       };
     }
   };
+
+Basic.parameters = { ...sharedArgActions };
 
 export const LongTitle = Basic.bind({});
 LongTitle.args = {
   title:
     'This is a long message. Text will remain on a single line and does not have a maximum width'
 };
+
+LongTitle.parameters = { ...sharedArgActions };
 
 export const WithButton = args =>
   class WithButton extends lng.Component {


packages/@lightningjs/ui-components/src/components/Tooltip/Tooltip.stories.js
```diff
@@ -58,24 +58,41 @@ export default {
   }
 };
 
-export const Basic = () =>
+const sharedArgActions = {
```
ImCoolNowRight, line 61: Can we change this to `sharedParameters` since the argActions are nested inside the object?

packages/@lightningjs/ui-components/src/components/Tooltip/Tooltip.stories.js
```diff
@@ -58,24 +58,41 @@ export default {
...
-          x: 1280 / 2
+          x: 1280 / 2,
+          delayVisible: args.delayVisible,
+          timeVisible: args.timeVisible
         }
       };
     }
   };
 
+Basic.parameters = { ...sharedArgActions };
```
ImCoolNowRight, line 86: Can we just assign the parameters directly instead of having to spread inside of an object?

packages/@lightningjs/ui-components/src/components/Tooltip/Tooltip.stories.js
```diff
@@ -58,24 +58,41 @@ export default {
...
+
 export const LongTitle = Basic.bind({});
 LongTitle.args = {
   title:
     'This is a long message. Text will remain on a single line and does not have a maximum width'
 };
 
+LongTitle.parameters = { ...sharedArgActions };
```
ImCoolNowRight, line 94: Same as above, can we just assign directly instead? Or do we want to leave it like this to make it clear that other properties can be added? I'm fine either way.

ImCoolNowRight (approved):
Looks good! Minor questions, but nothing functional.

packages/@lightningjs/ui-components/src/components/Tooltip/Tooltip.stories.js
soumyaloka: Yeah, as you said I use spread operator for that reason but I checked other components for consistency some of them use spread and some don't. I think either way is fine.

packages/@lightningjs/ui-components/src/components/Tooltip/Tooltip.stories.js
soumyaloka: Yes, I can change to sharedParameters, but crosschecked with other components, they are all using sharedArgActions? Either way is fine for me.

erautenberg: I think it's fine to keep it as arg actions to be consistent for now and we can always make a PR to update across the board later

anthony9187 (approved):

erautenberg (approved):

PR #133, opened 2023-03-07
test: add global image constructor

jazzyw:
## Description

- Add global image constructor to mock browser Image onload and onerror
- Removed all setTimeouts and txLoaded/txError emitters in Artwork

## References

Came from LUI-632 work (https://github.comcast.com/Lightning/lightning-ui/pull/1404#discussion_r1589803)

## Testing

Run `yarn test` and all the tests should pass. To remove the console.errors that appear in the terminal because of the brokenImage/image fail situation tests in Artwork, include `export CI=true;` in the `yarn test` command (run `export CI=true; yarn test`).

## Checklist


erautenberg (commented):
Not sure if it's a problem, but it seems odd to me that the texture h/w/x/y was updated to NaN in the snapshot 🤷‍♀️

erautenberg (commented):
This works, but I think the console errors are a little confusing. It makes me think the test failed. Is there another way to mock the error in the test env to make this clearer? Otherwise, no biggie.

packages/@lightningjs/ui-components/src/components/Artwork/Artwork.test.js
```diff
@@ -411,17 +417,18 @@ describe('Artwork', () => {
         src: sampleImg
       },
       {
+        // spyOnMethods: ['_resolveLoading', '_updateBlur']
```
jazzyw: Leaving a comment here to ask if anyone has an idea why if I include `_resolveLoading` in the `spyOnMethods` value array, the test ends up failing?

I was trying to add the SpyPromise methods to all the tests that recreate the artwork component but I noticed in this case, the test ends up failing with `_resolveLoading` defined and completely passes if it's not defined. In both cases though, I was console logging if the `resolveLoading` function was actually called and whether the method is defined or not, the function is still being called/ran through.

joshhowenstine: Hi @jazzyw , it looks like you need to await the update spy before testing the blur again. I believe that the blur promise resolves right after _resolveLoaded on the first cycle so you will need to wait for the next update cycle to run to get the results you are expecting. 

Screen Shot 2023-03-14 at 1 28 49 PM

soumyaloka (approved):

THoj13 (approved):

PR #134, opened 2023-03-07
fix(ListItemPicker): fix selection reset when switching modes

ImCoolNowRight:
## Description

This PR resolves an issue where the selection in ListItemPicker would be reset if the mode was changed. It also fixes a separate issue where the arrows would show up if the tone was changed and then the mode was set to unfocused/disabled.

## References

LUI-692

## Testing

Reset issue:

1. Navigate to ListItemPicker story.
2. Press right to get to either the "Description2" or "Description3" option.
3. Change the mode to unfocused or disabled.
4. The description text should remain and not be reset to "Description1" anymore.

Arrow issue:

1. Navigate to ListItemPicker story.
2. Press right to get to either the "Description1" or "Description3" option.
3. Change the tone to inverse and the mode to unfocused or disabled.
4. There should not be any arrows on the ListItemPicker.

## Automation

No story arguments are affected by these changes.

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/ListItem/ListItemPicker.js
+++ b/packages/@lightningjs/ui-components/src/components/ListItem/ListItemPicker.js
@@ -62,6 +62,7 @@ export default class ListItemPicker extends ListItem {
     this._updateArrowsAlpha();
     this._updateAlignment();
     this._previousMode = this.mode;
+    this._previousTone = this.tone;
   }
 
   _updateAlignment() {
@@ -149,7 +150,11 @@ export default class ListItemPicker extends ListItem {
       h: this.style.descriptionTextStyle.lineHeight,
       w
     };
-    if (this._optionsChanged || this._previousMode !== this.mode) {
+    if (
+      this._optionsChanged ||
+      this._previousMode !== this.mode ||
+      this._previousTone !== this.tone
+    ) {
       patchObject.items = this.options.map(option => ({
         type: Marquee,
         h: this.style.descriptionTextStyle.lineHeight,
--- a/packages/@lightningjs/ui-components/src/components/ListItem/ListItemPicker.js
+++ b/packages/@lightningjs/ui-components/src/components/ListItem/ListItemPicker.js
@@ -61,8 +61,6 @@ export default class ListItemPicker extends ListItem {
     this._updateArrows();
     this._updateArrowsAlpha();
     this._updateAlignment();
-    this._previousMode = this.mode;
-    this._previousTone = this.tone;
   }
 
   _updateAlignment() {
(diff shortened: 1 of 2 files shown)


soumyaloka (approved):
Looks good!

joshhowenstine:
This looks great, one small thing. Should we make sure the description text is a lighter value when tone is set to inverse? We can also handle this in a separate PR if more appropriate. 
Screen Shot 2023-03-08 at 9 14 49 AM

erautenberg:
> This looks great, one small thing. Should we make sure the description text is a lighter value when tone is set to inverse? We can also handle this in a separate PR if more appropriate. Screen Shot 2023-03-08 at 9 14 49 AM

It looks like this is a larger problem. When I toggle between neutral and inverse and focused/unfocused, the description seems to just be behind in its updates compared to the background and the title text. I don't think it should be a separate ticket though yet because it could be due to these changes to ensure the current selection is visible?

soumyaloka:
> > This looks great, one small thing. Should we make sure the description text is a lighter value when tone is set to inverse? We can also handle this in a separate PR if more appropriate. Screen Shot 2023-03-08 at 9 14 49 AM
> 
> It looks like this is a larger problem. When I toggle between neutral and inverse and focused/unfocused, the description seems to just be behind in its updates compared to the background and the title text. I don't think it should be a separate ticket though yet because it could be do to these changes to ensure the current selection is visible?

Yes sorry I didn't test this, but I see the issue now.
Screen Shot 2023-03-08 at 2 18 48 PM

ImCoolNowRight:
> > > This looks great, one small thing. Should we make sure the description text is a lighter value when tone is set to inverse? We can also handle this in a separate PR if more appropriate. Screen Shot 2023-03-08 at 9 14 49 AM
> > 
> > 
> > It looks like this is a larger problem. When I toggle between neutral and inverse and focused/unfocused, the description seems to just be behind in its updates compared to the background and the title text. I don't think it should be a separate ticket though yet because it could be do to these changes to ensure the current selection is visible?
> 
> Yes sorry I didn't test this, but I see the issue now. Screen Shot 2023-03-08 at 2 18 48 PM

Thanks all for catching this! Just needed to add a check to patch the items whenever the tone is changed, as well.

packages/@lightningjs/ui-components/src/components/ListItem/ListItemPicker.js
```diff
@@ -143,11 +145,17 @@ export default class ListItemPicker extends ListItem {
...
       visible: !this._collapse,
       h: this.style.descriptionTextStyle.lineHeight,
-      w,
-      items: this.options.map(option => ({
+      w
+    };
+    if (
+      this._optionsChanged ||
+      this._previousMode !== this.mode ||
+      this._previousTone !== this.tone
```
erautenberg: I'm not quite sure this new approach is correct. If the user patches in any style changes for the description, then this if statement will also get skipped. Yes, we need to patch a new set of items if the options array changes, but otherwise we probably should be updating the items directly to apply any style/theme updates (which would include changes from mode and tone).

ImCoolNowRight: You're right, I definitely overcomplicated it! I was able to revert this whole block back to what it originally was and just patch in the selected index to resolve the original issue.

erautenberg (approved):

PR #135, opened 2023-03-07
fix(withThemeStyles): add story documentation

ImCoolNowRight:
## Description

This PR updates the story for `withThemeStyles`, and adds documentation for it and the Theming Hierarchy page.

## References

LUI-648

## Testing

1. Go to the `withThemeStyles` story.
2. Ensure that the square is white when the `themeColor` toggle is true and blue when false.
3. Please read through the `withThemeStyles` docs and the `Theming Hierarchy` page to see if everything is covered and makes sense.

## Automation

There is a new story for the `withThemeStyles` mixin.

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/mixins/withThemeStyles/withThemeStyles.mdx
+++ b/packages/@lightningjs/ui-components/src/mixins/withThemeStyles/withThemeStyles.mdx
@@ -1,0 +1,139 @@
+<!--
+  Copyright 2023 Comcast Cable Communications Management, LLC
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+  SPDX-License-Identifier: Apache-2.0
+-->
+
+import { Canvas, Story } from '@storybook/addon-docs';
+# withThemeStyles
+
+Use `withThemeStyles` to create composable components that can be easily themed and decoupled from functionality.
+
+The `withThemeStyles` mixin provides flexibility with _composability_: the concept of combining the properties of multiple components to create a new component. This is not intended to be a replacement for building a component API, but rather a tool for augmentation.
+
+<Canvas>
+  <Story id="utilities-withthemestyles--with-theme-styles" />
+</Canvas>
+
+## Usage
+
+Let's use an example to see how we can use `withThemeStyles`. Let's say you have an unstyled `Box` component with a defined width and height:
+
+```js
+import lng from '@lightningjs/core';
+
+class Box extends lng.Component {
+  static _template() {
+    return {
+      w: 150,
+      h: 150,
+      rect: true
+    };
+  }
+}
+```
+
+If you want to re-use this `Box` to create a _new_ component called `BlueBox`, you can do that by extending the base class.
+
+```js
+class BlueBox extends Box {
+  static _template() {
+    return {
+      ...super._template(),
+      color: 0xff0000ff
+    };
+  }
+}
+```
+
+This is fine if you want a box that is always blue, but what if you want to change the color of your component without having to extend it and override that single value? Or if you want a box that changes based on certain factors, such as its [mode](?path=/docs/docs-theming-mode--page)? Or what if you want to have multiple color palettes (in LUI, these are called [tones](?path=/docs/docs-theming-tone--page)) that you can switch between depending on if your application is in light or dark mode?
+
+By utilizing `withThemeStyles`, you can tie the color of your box to the tokens used in your theme or a single style object. This can be done in the following ways.
+
+### Defining a Style Object
+
+The first (and suggested) way of handling styles is to wrap your component in the `withThemeStyles` mixin, and then define the `__themeStyle` getter.
+
+```js
+const style = {
+  color: 0xff0000ff
+};
+
+class Box extends withThemeStyles(lng.Component) {
+  static get __componentName() {
+    return 'Box';
+  }
+
+  static get __themeStyle() {
+    return style;
+  }
+
+  static _template() {
+    return {
+      w: 150,
+      h: 150,
+      rect: true
+    };
+  }
+
+  _update() {
+    this.color = this.style.color;
+  }
+}
+```
+
+With this setup, `Box` is able to have its color tied to a style value that exists outside of the `Box` class. This enables us to change the value of color externally, without the need to change any code in the `Box` class. Thus, the class can remain focused on the functional aspects of the component, and all style-related properties can be handled elsewhere.
+
+### Passing In a Style Object
+
+The second (and less preferred option) is to wrap your component in `withThemeStyles` and to pass in your style object as the second argument.
+
+```js
+const style = {
+  color: 0xff0000ff
+};
+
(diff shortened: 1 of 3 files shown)


packages/@lightningjs/ui-components/src/mixins/withThemeStyles/withThemeStyles.mdx
```diff
@@ -0,0 +1,139 @@
...
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+  SPDX-License-Identifier: Apache-2.0
+-->
+
+import { Canvas, Story } from '@storybook/addon-docs';
+# withThemeStyles
```
soumyaloka, line 21: Minor one, add space after line 19

packages/@lightningjs/ui-components/src/mixins/withThemeStyles/withThemeStyles.stories.js
```diff
@@ -0,0 +1,75 @@
+import lng from '@lightningjs/core';
```
soumyaloka, line 19: Just noticed, I think we should add copyright and license content before the imports like you had in mdx file.

packages/@lightningjs/ui-components/docs/ThemingHierarchy.stories.mdx
```diff
@@ -16,12 +16,29 @@
...
+
+A component's style file can consist of a base style object, a [tone style object](path=/docs/docs-theming-tone--page), and a [mode style object](path=/docs/docs-theming-mode--page). These three objects are deep merged, in that order, to create a single style object that serves as the foundation for the final component style object.
+
+## Theme's _componentConfig_
+
+Themes can have a `componentConfig` object that has nested component style objects, which serve to override any styles from component style files. The objects nested within `componentConfig` must have keys matching the `__componentName` of the target components. These are deep merged with the results from each of those files.
+
+## In-line Component Style Overrides
+
+When utilizing components within your application, you can use [withThemeStyles](path=/docs/utilities-withthemestyles--with-theme-styles) to override particular style values in just that use case. Those defined styles are merged with the result of the merge that happened above.
```
erautenberg: This sentence might be a little confusing
> Those defined styles are merged with the result of the merge that happened above.

Maybe something like this? Kind of confusing concept in general though
> Those custom styles are deep-merged with the resulting object from the previous steps.

packages/@lightningjs/ui-components/docs/ThemingHierarchy.stories.mdx
```diff
@@ -16,12 +16,29 @@
...
+
+## Extensions
+
+Any extensions that target your component are then merged with the result after in-line component style overrides to create the final component style object that will be referenced by your application's components.
```
erautenberg: Also a little wordy, maybe break it up?

Any extensions that target a component will be applied after the in-line component style overrides. This creates the final style object which will be referenced by the component via `this.style`.

packages/@lightningjs/ui-components/src/mixins/withThemeStyles/withThemeStyles.mdx
```diff
@@ -0,0 +1,140 @@
...
+
+The `withThemeStyles` mixin provides flexibility with _composability_: the concept of combining the properties of multiple components to create a new component. This is not intended to be a replacement for building a component API, but rather a tool for augmentation.
+
+<Canvas>
+  <Story id="utilities-withthemestyles--with-theme-styles" />
+</Canvas>
+
+## Usage
+
+Let's use an example to see how we can use `withThemeStyles`. Let's say you have an unstyled `Box` component with a defined width and height:
```
erautenberg: Maybe
> Here is an example of applying `withThemeStyles`. Start with an un-styled `Box` component containing a defined width and height.

packages/@lightningjs/ui-components/src/mixins/withThemeStyles/withThemeStyles.mdx
```diff
@@ -0,0 +1,140 @@
...
+    return {
+      w: 150,
+      h: 150,
+      rect: true
+    };
+  }
+}
+```
+
+If you want to re-use this `Box` to create a _new_ component called `BlueBox`, you can do that by extending the base class.
```
erautenberg: > In order to re-use this `Box` to create a _new_ component called `BlueBox`, we can extend the base class.

packages/@lightningjs/ui-components/src/mixins/withThemeStyles/withThemeStyles.mdx
```diff
@@ -0,0 +1,140 @@
...
+}
+```
+
+This is fine if you want a box that is always blue, but what if you want to change the color of your component without having to extend it and override that single value? Or if you want a box that changes based on certain factors, such as its [mode](?path=/docs/docs-theming-mode--page)? Or what if you want to have multiple color palettes (in LUI, these are called [tones](?path=/docs/docs-theming-tone--page)) that you can switch between depending on if your application is in light or dark mode?
+
+By utilizing `withThemeStyles`, you can tie the color of your box to the tokens used in your theme or a single style object. This can be done in the following ways.
+
+### Defining a Style Object
+
+The first (and suggested) way of handling styles is to wrap your component in the `withThemeStyles` mixin, and then define the `__themeStyle` getter.
```
erautenberg: I would argue that the "suggested" way is actually to extend our Base component, which uses withThemeStyles already. Maybe add a line about that in here somewhere?

packages/@lightningjs/ui-components/src/mixins/withThemeStyles/withThemeStyles.mdx
```diff
@@ -0,0 +1,140 @@
...
+    };
+  }
+
+  _update() {
+    this.color = this.style.color;
+  }
+}
+```
+
+With this setup, `Box` is able to have its color tied to a style value that exists outside of the `Box` class. This enables us to change the value of color externally, without the need to change any code in the `Box` class. Thus, the class can remain focused on the functional aspects of the component, and all style-related properties can be handled elsewhere.
```
erautenberg: Just simplifying some language "able to have its" -->
> With this setup, the `Box` component's color property is mapped to a style value external to the `Box` class.

packages/@lightningjs/ui-components/src/mixins/withThemeStyles/withThemeStyles.mdx
```diff
@@ -0,0 +1,140 @@
...
+
+### Passing In a Style Object
+
+The second (and less preferred option) is to wrap your component in `withThemeStyles` and to pass in your style object as the second argument.
```
erautenberg: maybe explain why one is preferred over another? for example, this is specific to in-line/component _instance_ updates, versus the first way is more reusable and creates a new component with a given style. The second way is sort of an anti-pattern as far as design systems go

packages/@lightningjs/ui-components/src/mixins/withThemeStyles/withThemeStyles.mdx
```diff
@@ -0,0 +1,140 @@
...
+class NewComponent extends lng.Component {
+  _template() {
+    Box: {
+      type: withThemeStyles(Box, style);
+    }
+  }
+}
+```
+
+This method is usually more appropriate for in-line styling, rather than defining the styles of a new component. This is because you rely on the component already being set up for use with theming, and this method will only apply the given styles to that particular instance of your component. (We advise changes to be made to all instances of a component when possible to ensure proper adherence to your design system.)
```
erautenberg: cool just reading this, maybe list some of this before the example above and pull in some of the other recommendations I added

soumyaloka (approved):

erautenberg (approved):

PR #136, opened 2023-03-07
fix(TextBox): ensure component is adhering to width

jazzyw:
## Description

A bug was encountered where just setting w on the `TextBox` component does not actually change the width of the component. Only after applying `wordWrapWidth` does the `TextBox` update its width.

This PR ensures that if a user defines the width (either with `w` or `wordWrapWidth` in `textStyles`) `wordWrapWidth` will be set on the `TextBox` component.

Updates to the `TextBox` documentation was also done in this PR in accordance to LUI-657.

## References

LUI-660 and LUI-657

## Testing

Go to the `TextBox` component

- Ensure initial width is 600
- In the TextBox story, remove `wordWrapWidth` from `textStyle` in the Basic story
- Add a w value and confirm width is adhering to the set w

For example:
```js
TextBox: {
  type: TextBox,
  style: { textStyle: { maxLines: 3 } },
  w: 400
}
```

- `TextBox`'s `InlineContent`'s w should not be changing/the entire component should be showing even when w is set to a small value (this work/issue will be addressed in another ticket)
- Modify and add spaces to the title in `Button` and there shouldn't be truncation on the text until fixed is set to true (test ex strings: "1234567898786756576788985678987 Button" and "To Be Removed")
- Modify and add spaces to the title in `Tooltip`, the spaces should not make new lines and the text should be continuous on 1 line

## Checklist


jazzyw:
Solution is in another branch --a new PR will be created for this work

PR #138, opened 2023-03-07
ci: workflow optimizations

joshhowenstine:
## Description

1. Runners should have the GITHUB_TOKEN env variable built in so GH_TOKEN is not required https://github.com/semantic-release/github

2. Pull requests should not require any permissions above read to run tests

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,7 +38,6 @@ jobs:
           yarn workspace @lightningjs/ui-components run semantic-release --no-ci -e semantic-release-monorepo 
           yarn workspace @lightningjs/ui-components-test-utils run semantic-release --no-ci -e semantic-release-monorepo
         env:
-          GH_TOKEN: ${{secrets.GH_TOKEN}}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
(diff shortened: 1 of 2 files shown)


erautenberg (approved):

soumyaloka (approved):

.github/workflows/release.yml
```diff
@@ -38,7 +38,6 @@ jobs:
           yarn workspace @lightningjs/ui-components run semantic-release --no-ci -e semantic-release-monorepo 
           yarn workspace @lightningjs/ui-components-test-utils run semantic-release --no-ci -e semantic-release-monorepo
         env:
-          GH_TOKEN: ${{secrets.GH_TOKEN}}
```
anthony9187, line 41: we dont need to replace this with another token?

joshhowenstine: From what I am reading a new token will be added to the workflow each time it runs. So I don't believe we will need to add it. https://docs.github.com/en/actions/security-guides/automatic-token-authentication

joshhowenstine: Oh, actually. I need to reference it through secrets. Will make that update.

anthony9187 (approved):

PR #142, opened 2023-03-09
fix(Input): update listening control in disabled mode

soumyaloka:
## Description

In this PR, I have fixed listening control issue in disabled mode.

## References

LUI-716

## Testing

Navigate to Input in disabled mode, try setting listening to true and make sure the cursor is not active in disabled mode and 
should be active only in focused and unfocused mode.

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Input/Input.mdx
+++ b/packages/@lightningjs/ui-components/src/components/Input/Input.mdx
@@ -61,7 +61,7 @@ class Example extends lng.Component {
 
 ### Setting/Inserting values
 
-"The `Input` component tracks its current value and cursor position, and provides different ways to manage the value state. When the listening mode is enabled, the cursor blinks as a `placeholder`, whether the component is in focused or unfocused mode."
+The `Input` component tracks its current value and cursor position, and provides different ways to manage the value state. When the listening mode is enabled, the cursor blinks as a `placeholder`, whether the component is in focused or unfocused mode.
 
 ```js
 this.tag('Input').value = 'Hello';
--- a/packages/@lightningjs/ui-components/src/components/Input/Input.mdx
+++ b/packages/@lightningjs/ui-components/src/components/Input/Input.mdx
@@ -61,7 +61,7 @@ class Example extends lng.Component {
 
 ### Setting/Inserting values
 
-The `Input` component keeps track of it's current `value` and cursor `position`. You can manage the value state in a few ways. Let's work off of the above example. We see the cursor blinking which is the `placeholder` value when listening mode is true and component in focused mode.
+The `Input` component keeps track of it's current `value` and cursor `position`. You can manage the value state in a few ways. Let's work off of the above example. We see the cursor blinking which is the `placeholder` value when listening mode is true and component in focused and unfocused mode.
 
 ```js
 this.tag('Input').value = 'Hello';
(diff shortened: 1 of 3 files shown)


erautenberg (commented):
This looks good to me! ~Just wondering if maybe as a separate ticket we should make the cursor blinking configurable in case a user doesn't want it at all?~ Nevermind, I assume that's what `cursorBlink` does?

arwehrman (approved):

packages/@lightningjs/ui-components/src/components/Input/Input.mdx
```diff
@@ -61,7 +61,7 @@ class Example extends lng.Component {
 
 ### Setting/Inserting values
 
-The `Input` component keeps track of it's current `value` and cursor `position`. You can manage the value state in a few ways. Let's work off of the above example. We see the cursor blinking which is the `placeholder` value when listening mode is true and component in focused mode.
+The `Input` component keeps track of it's current `value` and cursor `position`. You can manage the value state in a few ways. Let's work off of the above example. We see the cursor blinking which is the `placeholder` value when listening mode is true and component in focused and unfocused mode.
```
erautenberg: `it's` should be `its` in this case

erautenberg: Can we reword the second sentence a bit? It's a bit confusing on the second half

packages/@lightningjs/ui-components/src/components/Input/Input.mdx
```diff
@@ -102,6 +102,18 @@ this.tag('Input').value; //=> 'oo'
 this.tag('Input').position; //=> 0
 ```
 
+### listening
+
+When listening property is true, it enables cursor to blink which then allow user to enter text.
```
erautenberg: Maybe just
```When `listening` is true, the cursor begins blinking and the user is able to enter text.```

File: packages/@lightningjs/ui-components/src/components/Input/Input.mdx [hunk 3]
```diff
@@ -102,6 +102,18 @@ this.tag('Input').value; //=> 'oo'
...
+
+```js
+this.tag('Input').listening = true;
+When listening is true, the cursor blinks and allows user to enter text.
+
+this.tag('Input').listening = false;
+When listening is false, the cursor will not be visible and does not allow user to enter text.
```
erautenberg: this and line 111 should have `// ` in front so it looks like a code comment

packages/@lightningjs/ui-components/src/components/Input/Input.mdx
soumyaloka: Yes, I agree. I will update it.

soumyaloka: Something like this:
"The 'Input' component tracks its current value and cursor position, and provides different ways to manage the value state. When the listening mode is enabled, the cursor blinks as a placeholder, whether the component is in focused or unfocused mode."

packages/@lightningjs/ui-components/src/components/Input/Input.mdx
```diff
@@ -61,7 +61,7 @@ class Example extends lng.Component {
 
 ### Setting/Inserting values
 
-The `Input` component keeps track of it's current `value` and cursor `position`. You can manage the value state in a few ways. Let's work off of the above example. We see the cursor blinking which is the `placeholder` value when listening mode is true and component in focused mode.
+"The `Input` component tracks its current value and cursor position, and provides different ways to manage the value state. When the listening mode is enabled, the cursor blinks as a `placeholder`, whether the component is in focused or unfocused mode."
```
ImCoolNowRight: Why is this line in quotation marks?

ImCoolNowRight (approved):
Looks great! Just one small thing on the docs page.

packages/@lightningjs/ui-components/src/components/Input/Input.mdx
soumyaloka: oops that was unintentional. Thanks for catching. I will remove them.

PR #155, opened 2023-03-17
fix(TextBox): ensure component adheres to width

jazzyw:
## Description

A bug was encountered where just setting w on the `TextBox` component does not actually change the width of the component. Only after applying `wordWrapWidth` does the `TextBox` update its width.

This PR ensures that if a user defines the width (either with `w` or `wordWrapWidth` in `textStyles`) `wordWrapWidth` will be set on the `TextBox` component. This way the text will actually conform to the set width (even when no explicit `wordWrapWidth` is defined).

A `fixed` prop was added to `TextBox` similar to the `fixed` prop in `Button` that will need to be set to `true` to update the text width if the width of the text is set by providing a `w` value.

Updates to the `TextBox` documentation were also done in this PR in accordance to LUI-657.
Likewise a section on the updates to the width/different cases of setting the width were also added to the `TextBox` documentation.

## References

LUI-660 and LUI-657

## Testing

Go to the `TextBox` component

- Ensure initial width is 600
- Ensure fixed is set to true
- In the `TextBox` story, update the initial `w` value to a different value (eg 400) and confirm the width of the text updates to the set width value
- Remove the fixed prop setter and confirm that the text is no longer confirming to the set `w` value but should be following the rendered width
- In the `TextBox` story, remove the `w` value and add `wordWrapWidth` and a value to the `textStyle` object in the Basic story and confirm the width of the component reflects the set `wordWrapWidth` value

For example:
```js
TextBox: {
  type: TextBox,
  style: { textStyle: { wordWrapWidth: 500, maxLines: 3 } },
}
```

- `TextBox`'s `InlineContent`'s w should not be changing/the entire component should be showing even when w is set to a small value (this work/issue will be addressed in another ticket)
- Modify and add spaces to the title in `Button` and there shouldn't be truncation on the text until fixed is set to true (test ex strings: "1234567898786756576788985678987 Button" and "To Be Removed")
- Modify and add spaces to the title in `Tooltip`, the spaces should not make new lines and the text should be continuous on 1 line (eg a title like 'Longer Tooltip' should all be on 1 line and not on 2 lines)

**For all components that utilize `TextBox` (`Button`, `Tooltip`, etc), ensure the text is centered in the middle of the component.

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/TextBox/TextBox.stories.js
+++ b/packages/@lightningjs/ui-components/src/components/TextBox/TextBox.stories.js
@@ -45,6 +45,7 @@ export const Basic = () =>
       return {
         TextBox: {
           type: TextBox,
+          fixed: true,
           w: 600,
           style: { textStyle: { maxLines: 3 } }
         }
--- a/packages/@lightningjs/ui-components/src/components/TextBox/TextBox.stories.js
+++ b/packages/@lightningjs/ui-components/src/components/TextBox/TextBox.stories.js
@@ -55,7 +55,9 @@ export const Basic = () =>
 
 Basic.args = {
   content: lorum,
-  marquee: false
+  marquee: false,
+  fixed: true,
+  w: 600
 };
 
 Basic.argTypes = {
@@ -72,6 +74,22 @@ Basic.argTypes = {
       "Scrolls text when a single line of text is longer than the textbox's width",
     table: {
       defaultValue: { summary: false }
+    }
+  },
+  fixed: {
+    control: 'boolean',
+    description:
+      'Flag that when set to `true`, allows the width of the component to be set with `w`',
+    table: {
+      defaultValue: { summary: 'undefined' }
+    }
+  },
+  w: {
+    control: 'number',
+    description:
+      'When the `fixed` property is `true`, this will set the width of the component',
+    table: {
+      defaultValue: { summary: 0 }
     }
   }
 };
--- a/packages/@lightningjs/ui-components/src/components/TextBox/TextBox.stories.js
+++ b/packages/@lightningjs/ui-components/src/components/TextBox/TextBox.stories.js
@@ -87,7 +87,7 @@ Basic.argTypes = {
   w: {
     control: 'number',
     description:
-      'When the `fixed` property is `true`, this will set the width of the component',
+      'When the `fixed` property is `true`, this will set the width of the component (but only if `wordWrapWidth` is not defined otherwise the width will reflect the `wordWrapWidth` value)',
     table: {
       defaultValue: { summary: 0 }
     }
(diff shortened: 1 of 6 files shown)


erautenberg (commented):
Could we add a control to the TextBox story for `fixed`?

Also, just confirming, if I _don't_ set `fixed` in the story and I set a `w` but not a `wordWrapWidth`, then the set `w` is not used for anything, it just uses the render width of the component. So basically, if you set a `w`, you should set `fixed` as well, right?

packages/@lightningjs/ui-components/src/components/TextBox/TextBox.stories.js
```diff
@@ -71,6 +75,22 @@ Basic.argTypes = {
...
+    description:
+      'Flag that when set to `true`, allows the width of the component to be set with `w`',
+    table: {
+      defaultValue: { summary: 'undefined' }
+    }
+  },
+  w: {
+    control: 'number',
+    description:
+      'When the `fixed` property is `true`, this will set the width of the component',
```
jazzyw: Should we specify here that, it will only set the width to the `w` value (assuming `fixed` is `true`) if `wordWrapWidth` is not defined?

**If `wordWrapWidth` is defined at all, the width will reflect the `wordWrapWidth` even if a `w` value is provided and `fixed` is set to `true`

erautenberg: agreed

erautenberg (approved):

packages/@lightningjs/ui-components/src/components/TextBox/TextBox.stories.js
```diff
@@ -71,6 +75,22 @@ Basic.argTypes = {
     table: {
       defaultValue: { summary: false }
     }
+  },
+  fixed: {
+    control: 'boolean',
+    description:
+      'Flag that when set to `true`, allows the width of the component to be set with `w`',
+    table: {
+      defaultValue: { summary: 'undefined' }
```
soumyaloka: Minor one, I think we should change defaultValue for fixed control to boolean.

soumyaloka (approved):

ChrisArasin (approved):
Works as I'd expect!

packages/@lightningjs/ui-components/src/components/TextBox/TextBox.stories.js
jazzyw: If we don't set `fixed`, it technically is 'undefined' (console logging shows 'undefined') so that's why I put 'undefined' as the default. However, since it is supposed to be a boolean, is it better to show the default as 'false'?

soumyaloka: You are right , but to be consistent I would expect to show the default value as false here.
I think in the Button component for fixed control the default value is false.

PR #158, opened 2023-03-17
feat(withLightning): allow configuring story to remount on prop changes

THoj13:
## Description
Updates withLightning to allow configuring a story to remount the component when specified args or all args on a story change.
### **Remount from only specific properties changing:** 
- add `remount: true` as a property to the argType of property
```js
SomeComponent.argTypes = {
scrollIndex: {
  control: { type: 'number', min: 0 },
  description: '...',
  table: { defaultValue: { summary: 0 } },
  remount: true // remount the story component when scrollIndex changes
}
```

### **Remount for all properties changes:** 
- add `remountAll` true to the `parameters object. `mode` will be excluded from the properties that trigger remounting.
```js
OtherComponent.parameters = {
  remountAll: true
}
```

## Testing
**parameters.remountAll has been set to `true` in `Row.stories.js`**
In all Row stories:
1. navigate to any item on the Row besides the first (which is the default)
2. change the value of any controls
ER: the Row should remount. This can be observed by the seeing the component reload into view, which will reset the selected item in the row to the first item

**parameters.remountAll has been set to `true` in `TitleRow.stories.js`**
In all TitleRow stories
1. navigate to any item on the TitleRow besides the first (which is the default)
2. change the value of any control, excluding `title`
ER: the TitleRow should remount. This can be observed by the seeing the component reload into view, which will reset the selected item in the row to the first item
3. repeat step 1, then change only the `title` to a new value
ER: the TitleRow should not remount. The item that was in focus after step 3 should remain in focus and not remount with the first item selected until a different control changes.

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Column/Column.stories.js
+++ b/packages/@lightningjs/ui-components/src/components/Column/Column.stories.js
@@ -31,8 +31,7 @@ export default {
   parameters: {
     docs: {
       page: mdx
-    },
-    remountAll: true
+    }
   }
 };
 
--- a/packages/@lightningjs/ui-components/src/components/Column/Column.stories.js
+++ b/packages/@lightningjs/ui-components/src/components/Column/Column.stories.js
@@ -31,7 +31,8 @@ export default {
   parameters: {
     docs: {
       page: mdx
-    }
+    },
+    remountAll: true
   }
 };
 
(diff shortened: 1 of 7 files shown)


THoj13:
This broke at some point in the last few commits. Marking as a draft until fixed.

erautenberg (commented):
Looks great! I think it's unrelated to this, but when I'm on the Row story, if I toggle `alwaysScroll` and `lazyScroll` both to true, I get some weird behavior when trying to scroll all the way back from the right. I'm assuming there is a check missing somewhere to only do one or the other. I think the hierarchy should be:
- alwaysScroll
- neverScroll
- lazyScroll

though I'm open to changes on those. Can you just confirm this isn't a side effect of this PR and if not, let's make a new ticket for it?

arwehrman (approved):

THoj13:
@erautenberg 
> Can you just confirm this isn't a side effect of this PR and if not, let's make a new ticket for it?

This issue exists on develop as well, so ya it's probably using a combination both `lazyScroll` and `alwaysScroll` logic instead of just one. I'll submit for a ticket for this.

erautenberg (approved):

PR #159, opened 2023-03-21
fix(Tile): resolve last letter removal issue in Tile metadata

soumyaloka:
## Description

In this PR updated:
- In TextBox `this._updateMarquee() `is moved before the patches to Text and InlineContent to undefined. That change will now allow TextBox to update the Marquee component within it when content has been removed.
- In Marquee updated the getter for `Marquee.textContent` to only return strings and default to an empty string

## References

LUI-723

## Testing

Navigate to the Tile component  in the Storybook.
Try removing the Metadata Title or Description text and make sure it removes the entire text in canvas if you remove the entire text in controls.

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Marquee/Marquee.js
+++ b/packages/@lightningjs/ui-components/src/components/Marquee/Marquee.js
@@ -267,7 +267,15 @@ export default class Marquee extends Base {
   }
 
   get textContent() {
-    return this.title && this.title.text ? this.title.text : this.title;
+    if (this.title) {
+      if (typeof this.title === 'string') {
+        return this.title;
+      }
+      if (typeof this.title.text === 'string') {
+        return this.title.text;
+      }
+    }
+    return '';
   }
 
   get _loopWidth() {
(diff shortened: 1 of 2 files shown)


packages/@lightningjs/ui-components/src/components/Marquee/Marquee.js
```diff
@@ -267,7 +267,15 @@ export default class Marquee extends Base {
   }
 
   get textContent() {
-    return this.title && this.title.text ? this.title.text : this.title;
```
anthony9187, line 270: instead of adding nested type checks, we can do this with just an `||`. we could also probably use a `?.` here instead of the ternary

soumyaloka: How about something like this: `return this.title?.text ?? this.title ?? '';`

erautenberg: You could probably even just do `this.title?.text || this.title` unless you absolutely need the empty string and not undefined/null.

soumyaloka: Yes, I need it to return empty string.

erautenberg (approved):

ImCoolNowRight (approved):

PR #165, opened 2023-03-23
fix(TabBar): adds column container story for scrolling tab content

arwehrman:
## Description

This PR adds a story in TabBar to show how adding a container helps expected scroll behavior when  TabBar content has multiple rows

## References

LUI-717

## Testing

- clone branch
- go to Tabar Multiple Rows story 
- navigate to Tab 1
- press down arrow to set focus on first row
- continue pressing down, former rows should not be visiable
 

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/TabBar/TabBar.stories.js
+++ b/packages/@lightningjs/ui-components/src/components/TabBar/TabBar.stories.js
@@ -316,6 +316,14 @@ class ColumnContainer extends Base {
   _getFocused() {
     return this._Column;
   }
+
+  _update() {
+    //super.update();
+    this._updateTilePadding();
+  }
+  _updateTilePadding() {
+    console.log(this);
+  }
 }
 
 export const MultipleRows = () =>
--- a/packages/@lightningjs/ui-components/src/components/TabBar/TabBar.stories.js
+++ b/packages/@lightningjs/ui-components/src/components/TabBar/TabBar.stories.js
@@ -283,14 +283,14 @@ class ColumnContainer extends Base {
     return {
       ClippingOffsetContainer: {
         w: 1920 - 160,
-        h: 500,
+        h: 600,
         clipping: true,
-        y: context.theme.layout.gutterY.sm * -1,
-        x: context.theme.layout.marginX * -1,
+        y: context.theme.layout.gutterY.xs * -1,
+        x: 40 * -1,
         Column: {
           type: Column,
-          y: context.theme.layout.gutterY.sm,
-          x: context.theme.layout.marginX,
+          y: 30,
+          x: 40,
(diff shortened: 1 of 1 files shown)


packages/@lightningjs/ui-components/src/components/TabBar/TabBar.stories.js
THoj13, line 1: This looks like a good solution to me. I think this scrolling logic is outside the scope of what TabBar is responsible for handling. For this iteration, it could be helpful to add documentation/recipe for how to render a column as tabContent like this. I'm not sure if it'd be best to include that in the TabBar or Column docs or both. If you see value in adding text on the story as well, `parameters.storyDetails` is available.

THoj13 (approved):

erautenberg (commented):
So I think this is a good first pass, but because of the way the clipping is set up, the focus state of the tile is cut off. It looks like things are slightly offset on the SUITE side and need to be adjusted, but basically what we do is add extra padding to cover that scale up and then subtract that amount from the Items container inside the Column to adjust it all back into place.

packages/@lightningjs/ui-components/src/components/TabBar/TabBar.stories.js
```diff
@@ -267,3 +268,90 @@ CustomTabs.args = {
...
+  }
+}
+
+export const MultipleRows = () =>
+  class MultipleRows extends lng.Component {
+    static _template() {
+      return {
+        TabBar: {
+          type: TabBarComponent,
+          w: 1920 - 160,
```
erautenberg: you should actually be able to reference `context.theme.layout.marginX * 2` here. Or even `getColumnWidth(1)`

erautenberg:
This is closer, but the Y position is still off. If you look at the picture, the scaled up top of the tile get cuts off. The left side is good though!
Screen Shot 2023-04-11 at 4 01 00 PM

packages/@lightningjs/ui-components/src/components/TabBar/TabBar.stories.js
arwehrman: couldn't find `getColumnWidth` in the utils, so used `context.theme.layout.marginX * 2`

erautenberg:
Much better, thank you! Only weird thing I'm seeing now is that the TabBar itself scrolls. Can we make sure it has the right width on it so that doesn't happen? Or an `autoResizeWidth: true`?

arwehrman:
> Much better, thank you! Only weird thing I'm seeing now is that the TabBar itself scrolls. Can we make sure it has the right width on it so that doesn't happen? Or an `autoResizeWidth: true`?
This is fixed now

erautenberg (approved):

PR #168, opened 2023-03-27
fix(Button): update content width

salhan12:
## Description

The text was sliding from right to left instead of being mounted. Updated Button text sliding to be mounted. 

## References

LUI-674

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Button/Button.js
+++ b/packages/@lightningjs/ui-components/src/components/Button/Button.js
@@ -193,7 +193,7 @@ export default class Button extends Surface {
 
     // reference transition targets to ensure we aren't checking against intermittent values
     if (this._Content.transition('w').targetValue !== this._contentW) {
-      contentDimensionsPatch.w = this._contentW;
+      this._Content.w = this._contentW;
     }
 
     if (this._Content.y !== y) {


arwehrman (approved):

ImCoolNowRight (approved):

erautenberg (changes requested):
I'd like to wait to merge and release this so that I can test it a little more. The structure of this method seems really odd to me with this change and I want to make sure there are no side effects of never animation the width change. Also, it doesn't seem like an issue that was introduced in this PR, but the text on the Button doesn't seem quite centered, it seems like it always leans to the left.

anthony9187:
> I'd like to wait to merge and release this so that I can test it a little more. The structure of this method seems really odd to me with this change and I want to make sure there are no side effects of never animation the width change. Also, it doesn't seem like an issue that was introduced in this PR, but the text on the Button doesn't seem quite centered, it seems like it always leans to the left.

+1 to this, if the goal is to prevent animating in some cases we should rework this method to be a little more clear about the intention. with these changes we could wind up in a situation where the short circuit prevents width from being animated but then the height still gets animated and things could be weird

ChrisArasin:
I think there's a reasonable argument that since the underlying text (and icon) elements are always "snapping" to their new sizes in a texture update, while x/y positioning may want to be smoothed, the width can always just snap to the proper size. This way, if someone updates button text to be much shorter or much longer, it is instantly positioned correctly, rather than seeing the text update with the previous position then slide into place.

ChrisArasin:
I think there also may be a decent chance you actually wouldn't want the y value (`this._Content.y`) to smooth either and just doing a patch for both would be fine... Trying to think of a time you'd want that update to smooth.... If someone patches a new height on the button, you'd want a patched update. If someone smooths the height on the button, the update methods are called on each tick so you would see this smooth to match anyway.

ChrisArasin:
Tested on a couple apps and definitely makes things look better as is.

erautenberg (approved):

PR #173, opened 2023-04-03
chore: update Row, Column and Navigation Manager docs

soumyaloka:
## Description

In this PR, styles table has been updated and also improved the description of controls for Row and Column components.

## References

LUI-748

## Checklist


packages/@lightningjs/ui-components/src/components/Column/Column.mdx
```diff
@@ -57,7 +53,7 @@ class BasicExample extends lng.Component {
 }
 ```
 
-The `Column` component extends [FocusManager][fm-docs], so you need to make sure that `_getFocused` returns the `Column` component.
+The `Column` component extends FocusManager, so you need to make sure that `_getFocused` returns the `Column` component.
```
erautenberg: Could you add a link to the FocusManager here? I assume that's what "fm-docs" was supposed to be.

erautenberg (approved):
I think we can set the `type` of `itemTransition` in Row/Column to be `Object` instead of just empty.

soumyaloka:
> I think we can set the `type` of `itemTransition` in Row/Column to be `Object` instead of just empty.

`itemTransition`  type is `<DocsLink id="lng.Transition" />` now , I can replace that with object.

packages/@lightningjs/ui-components/src/components/Column/Column.mdx
ImCoolNowRight: Do we still want to add this link in?

arwehrman (approved):

ImCoolNowRight (approved):
Looks good! Just left a comment about the link to the FocusManager if we still want it added.

PR #178, opened 2023-04-07
fix(Input): add value getter and setter to match documentation

### LINKED ISSUE #176: Missing type definitions for Input field methods and Tab component not exported
### Description
Firstly, I'd like to express my appreciation for the effort put into this package. It has been a significant help in creating lightning UI.

However, I have encountered some inconsistencies during implementation:

1. The Input field is missing type definitions for the insert(x: string) and backspace() methods. These methods are functional but not defined in the type spec of the input.
2. The documentation includes information about setting the value directly using input.value = "newvalue", but this does not work, regardless of the listening setting.
3. The Tab component does not appear to be exported from the package. I am only able to find the TabBar component.

Thank you again for the great work, and I look forward to your response. Have a wonderful day!

THoj13:
## Description
Setting the text contents of the Input via the `value` property has been in the Input documentation but was not implemented. This change adds a setter and getter for the `value` property.  In addition, I revised the Input documentation on the various methods to provide more details and use consistent verbiage.

## References
Addresses issue 2 of 3 in https://github.com/rdkcentral/Lightning-UI-Components/issues/176
>The documentation includes information about setting the value directly using input.value = "newvalue", but this does not work, regardless of the listening setting.

## Testing
I'm not sure there needs to be story for updating the Input with these setters. A unit test has been added to test this behavior, so running all unit tests should pass.

## Automation
N/A

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Input/Input.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/Input/Input.d.ts
@@ -39,9 +39,8 @@ export default class Input extends Button {
   mask?: string;
   password?: boolean;
   position?: number;
+  value?: string;
   get isCursorActive(): boolean;
-  set value(value: string);
-  get value(): string;
   get style(): InputStyle;
   set style(v: StylePartial<InputStyle>);
 
--- a/packages/@lightningjs/ui-components/src/components/Input/Input.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/Input/Input.d.ts
@@ -40,6 +40,8 @@ export default class Input extends Button {
   password?: boolean;
   position?: number;
   get isCursorActive(): boolean;
+  set value(value: string);
+  get value(): string;
   get style(): InputStyle;
   set style(v: StylePartial<InputStyle>);
 
--- a/packages/@lightningjs/ui-components/src/components/Input/Input.mdx
+++ b/packages/@lightningjs/ui-components/src/components/Input/Input.mdx
@@ -71,7 +71,7 @@ The cursor position can be manually updated by setting the `position` property. 
 // dislay Hello in the input
 this.tag('Input').value = 'Hello';
 
-// update the cursor position in the input from index 0 (before the H in Hellow) to index 5 (after the o in Hello)
+// update the cursor position in the input from index 0 (before the H in Hello) to index 5 (after the o in Hello)
 this.tag('Input').position = 5;
 ```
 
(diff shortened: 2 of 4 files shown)


packages/@lightningjs/ui-components/src/components/Input/Input.mdx
```diff
@@ -61,19 +61,22 @@ class Example extends lng.Component {
 
 ### Setting/Inserting values
 
-The `Input` component tracks its current value and cursor position, and provides different ways to manage the value state. When the listening mode is enabled, the cursor blinks as a `placeholder`, whether the component is in focused or unfocused mode.
+The `Input` component provides different ways to manage the input text and cursor position. When the `listening` property is enabled, the cursor blinks as a placeholder, whether the component is in focused or unfocused mode.
+
+The text displayed in the `Input` can can be directly set via the `value` property. The following example will display the input with the text "Hello".
```
ImCoolNowRight: You have "can can be directly"

packages/@lightningjs/ui-components/src/components/Input/Input.mdx
```diff
@@ -85,13 +88,15 @@ this.tag('Input').position = 5;
...
-To clear the current value, use `clear()`. This also repositions the cursor.
+To clear the current value use `clear()`. This also repositions the cursor to index `0`.
+The `Input` must be in focused or unfocused mode and the `listening` property enabled to use `clear()`.
 
 ```js
 this.tag('Input').clear();
 ```
 
-Use the `backspace()` method to handle insert deletions. This will remove the character at the current cursor `position - 1` and moves the cursor back. To illustrate this behavior, let's look at a simple example:
+To handle insert deletions use the `backspace()` method. This will remove the character at the current cursor `position - 1` and moves the cursor back.
```
ImCoolNowRight: "handle insert deletions" sounds confusing to me. Can we rephrase this?

ImCoolNowRight (approved):
Two minor wording comments in the docs, but functionally it looks great!

packages/@lightningjs/ui-components/src/components/Input/Input.d.ts
```diff
@@ -40,6 +40,8 @@ export default class Input extends Button {
   password?: boolean;
   position?: number;
   get isCursorActive(): boolean;
+  set value(value: string);
+  get value(): string;
```
erautenberg: does this need to be separated out or can it just be `value: string` like the other props above it?

THoj13: I'm not sure, I assumed it would match how the other getters/setters are typed. I'll see if I can find TS docs on this.

THoj13: The TS sandbox generates the `.d.ts` file like above. I tested both ways in a TS app and it the compiler accepts either. I'll update it to the shorter version.

erautenberg (approved):

PR #185, opened 2023-04-12
fix(Tab): make index.js exports match index.d.ts exports

erautenberg:
## Description

Making the index.js and index.d.ts files better match for `components/index.js` and `TabBar/index.js`

## Checklist


soumyaloka:
Emily, I noticed most of the import statements in `index.js` include the .js extension, while others do not. For the consistency, I suggest we drop the .js extension from all import statements or vice versa which ever works for you.

packages/@lightningjs/ui-components/src/components/TabBar/index.js
```diff
@@ -16,4 +16,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-export { default as default } from './TabBar.js';
+import TabBar from './TabBar';
+import Tab from './Tab';
+
+export { TabBar as default, Tab };
```
THoj13, line 22: TabBar.js also exports Tab. That can probably be removed with this change, but also won't break anything if it remains there too.

THoj13 (approved):
I left a comment about another optional change, but this looks good to me.

packages/@lightningjs/ui-components/src/components/TabBar/index.js
anthony9187: I think it'd be best to remove it, this is how we wind up with weird circular dependencies 😬

erautenberg: Oh I didn't realize that 🤔 it seems a little odd to me. Thoughts on removing that and just having it live here in the index file?

anthony9187: I think it makes the most sense to remove the Tab export from TabBar and have index.js handle all the exports of the module

anthony9187 (approved):

PR #187, opened 2023-04-14
chore(Storybook): update default control values

soumyaloka:
## Description

In this PR, I have updated the default value for the control and updated canvas path to make it visible in FocusManager and NavigationManager documentation.

## References

LUI-754

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Keyboard/Keyboard.stories.js
+++ b/packages/@lightningjs/ui-components/src/components/Keyboard/Keyboard.stories.js
@@ -113,7 +113,7 @@ Keyboard.argTypes = {
     description: "Center the keys within it's set width of keyboard",
     control: 'boolean',
     table: {
-      defaultValue: { summary: true }
+      defaultValue: { summary: false }
     }
   }
 };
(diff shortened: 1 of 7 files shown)


arwehrman (approved):

packages/@lightningjs/ui-components/src/components/Keyboard/Keyboard.stories.js
```diff
@@ -113,7 +113,7 @@ Keyboard.argTypes = {
     description: "Center the keys within it's set width of keyboard",
     control: 'boolean',
     table: {
-      defaultValue: { summary: false }
+      defaultValue: { summary: true }
```
THoj13: The default value for `Keyboard.centerKeys` is `undefined` in the `Keyboard` component. I think `false` would be the more accurate value here.

packages/@lightningjs/ui-components/src/components/Marquee/Marquee.stories.js
```diff
@@ -69,7 +69,7 @@ Marquee.argTypes = {
     control: 'boolean',
     description: 'Start scrolling on initiation',
     remount: true,
-    table: { defaultValue: { summary: false } }
+    table: { defaultValue: { summary: true } }
```
THoj13: `Marquee.autoStart` by default is `false`. 
[This is where it is where that value is defaulted in the component](https://github.com/rdkcentral/Lightning-UI-Components/blob/develop/packages/%40lightningjs/ui-components/src/components/Marquee/Marquee.js#L79)

THoj13 (changes requested):
I believe the changes to defaultValue in the Keyboard and Marquee stories files should be reverted. The defaultValue in that table should reflect what the default value of that prop is in the component, not the initial value passed to it in the story.

THoj13 (approved):

PR #199, opened 2023-05-03
ci: swap to PAT (GH_TOKEN) for semantic release to get around branch rules

erautenberg:
## Description

Currently, the semantic-release-bot uses the GITHUB_TOKEN instead of the GH_TOKEN, which is our service account's PAT and will have permissions to edit the branches.

@HarryLinux is going to look into whether we need `persist-credential: false`:
https://github.com/semantic-release/semantic-release/discussions/2557

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -2,7 +2,8 @@ name: Pull Request
 
 on:
   pull_request:
-    types: [opened, edited]
+    types: [opened, edited, reopened, synchronize, ready_for_review]
+
 permissions: read-all
 jobs:
   quality:
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -6,5 +6,4 @@ on:
 permissions: read-all
 jobs:
   quality:
-    name: Quality
     uses: ./.github/workflows/quality.yml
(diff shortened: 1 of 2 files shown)


ImCoolNowRight (approved):

HarryLinux (changes requested):
from what I've been reading in [this discussion thread](https://github.com/semantic-release/semantic-release/discussions/2557), it seems we need to add `persist-credentials: false` to L23 under the checkout action `with` given that the token we're using contains permissions to push to a protected branch. Setting persist-credentials to false apparently instructs GitHub to use the token when pushing to the branch

erautenberg:
@HarryLinux thanks, updated!

HarryLinux (approved):
lgtm

.github/workflows/pull-request.yml
```diff
@@ -1,6 +1,8 @@
 name: Pull Request
 
-on: [pull_request]
+on:
+  pull_request:
+    types: [opened, edited]
 permissions: read-all
 jobs:
   quality:
```
ImCoolNowRight, line 9: I believe this is where @HarryLinux suggested we add `uses: actions/checkout@v3` to checkout the repo before we run the Quality workflow.

PR #200, opened 2023-05-03
chore(NavigationManager): reduce number of update calls from onAfterU…

THoj13:
## Description
When items are added to NavigationManager, a function is assigned to an `onAfterUpdate` property on each item. Each function triggers an update if one of the specified sizing or positioning properties ([defined by the watchProps variable](https://github.com/rdkcentral/Lightning-UI-Components/pull/200/files#diff-4cf29152f6e19e45ffb9a2a5c968d8a4857ad5d4c413c184e9304e59efe050f8R278-R284)) changes.

### Related Pull Request
A follow up PR, https://github.com/rdkcentral/Lightning-UI-Components/pull/222,  has been created to abstract the onAfterUpdate logic used in NavigationManager. _withAfterUpdate to a utility function that can be reused in other components.

### NavigationManager
 The changes in this PR are intended to reduce the number of updates triggered from this process. The approaches taken were:
- remove the property associated with the main axis of the NavigationManager subclass from the list of properties evaluated for changes to trigger an update(ex. `x` for `Row`, `y` for `Column`). These properties are updated when the component updates (via `NavigationManager._updateLayout`). Each item's main axis property would then be updated with a new value, triggering that item's `onAfterUpdate` function to run and run through another update cycle.
- if a property is transitioning to its next value, only trigger an update from it after that transition has completed. Prior to this change, `_update` was triggered multiple times as the value of the property transitioned to its target value.
- removed a forced stage update (`this.stage.update()`) from the appendItem method. I'm not sure the history of why that was initially needed, but removing it does not appear to break anything and it seems like the component continues to render correctly.

### Button:
- Removing the stage update from `NavigationManager.appendItems` caused the `KeyboardInput` story to stop rendering and throw an error when attempting to patch `x` to `this._Prefix` or `this._Suffix` in `Button._updatePositions`. Neither of those tags were defined, because `this.prefix` and `this.suffix` are set to `[]` in the KeyboardInputStories.
- Two getters, for `_hasPrefix` and `_hasSuffix` were previously evaluating to truthy values if an empty array or object was passed to `Button.prefix` or `Button.suffix`. This PR updates those getters to always return a boolean and return `false` if an empty array or object is set on those properties.

## References
LUI-797

## Testing
Compare stories in the following components to those on the deployed storybook instance. No visible changes between the branches should be present: 
- Row 
- Column
- Button
- Control (subclass of button)
- ControlRow (utilizes both NavigationManager and Control)

## Automation
n/a

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/NavigationManager/NavigationManager.js
+++ b/packages/@lightningjs/ui-components/src/components/NavigationManager/NavigationManager.js
@@ -349,6 +349,7 @@ export default class NavigationManager extends FocusManager {
     items.forEach(item => this._appendItem(item));
 
     this.queueRequestUpdate();
+    this._refocus();
   }
 
   appendItemsAt(items = [], idx) {
--- a/packages/@lightningjs/ui-components/src/components/NavigationManager/NavigationManager.js
+++ b/packages/@lightningjs/ui-components/src/components/NavigationManager/NavigationManager.js
@@ -271,6 +271,8 @@ export default class NavigationManager extends FocusManager {
   }
 
   _withAfterUpdate(component) {
+    const initialOnAfterUpdate = component.__core?._onAfterUpdate;
+
     component.onAfterUpdate = function (element) {
       let hasChanged = false;
       const watchProps = [
@@ -297,6 +299,11 @@ export default class NavigationManager extends FocusManager {
 
       if (hasChanged) {
         this.queueRequestUpdate();
+      }
+
+      // if a component already has an onAfterUpdate function, preserve that behavior
+      if (initialOnAfterUpdate) {
+        initialOnAfterUpdate(element);
       }
     }.bind(this);
 
--- a/packages/@lightningjs/ui-components/src/components/NavigationManager/NavigationManager.js
+++ b/packages/@lightningjs/ui-components/src/components/NavigationManager/NavigationManager.js
@@ -17,7 +17,7 @@
  */
 
 import FocusManager from '../FocusManager';
-import { getX, getY, getH, getW, max } from '../../utils';
+import { getX, getY, getH, getW, max, watchForUpdates } from '../../utils';
 import * as styles from './NavigationManager.styles';
 
 const directionPropNames = {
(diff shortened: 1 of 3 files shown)


packages/@lightningjs/ui-components/src/components/NavigationManager/NavigationManager.js
```diff
@@ -273,12 +273,28 @@ export default class NavigationManager extends FocusManager {
...
+        const nextValue = propTransition
+          ? propTransition.targetValue
+          : element[prop];
+
+        if (propTransition.isRunning()) {
+          return;
+        }
+
+        if (nextValue !== element[prevValueKey]) {
+          element[prevValueKey] = nextValue;
```
erautenberg: do you think any of this can/should be lifted out into a util we utilize later on?

erautenberg (commented):
On the Expandable Height Rows story of Column, sometimes the y position doesn't change on the first load:
http://localhost:8000/?path=/story/navigation-column--expandable-height-rows&globals=stageColor:false

packages/@lightningjs/ui-components/src/components/NavigationManager/NavigationManager.js
THoj13: I think that could be useful. I can move this to utils.

THoj13: I'll do that on a separate branch. I think doing that abstraction here could make reviewing the changes in this diff more difficult to see.

THoj13: Here's the PR that abstracts this: https://github.com/rdkcentral/Lightning-UI-Components/pull/222

anthony9187 (approved):
looks great! would the plan be to merge 222 into this one and then this into develop, or merge this then 222?

THoj13:
> looks great! would the plan be to merge 222 into this one and then this into develop, or merge this then 222?

Thanks! I was thinking merging 222 into this branch, then both into develop.

THoj13:
> On the Expandable Height Rows story of Column, sometimes the y position doesn't change on the first load: http://localhost:8000/?path=/story/navigation-column--expandable-height-rows&globals=stageColor:false

Thanks for catching this. It should be resolved now.

erautenberg (approved):

PR #202, opened 2023-05-04
fix(Announcer): Ignore empty speech strings

soumyaloka:
## Description

Refactored the code to explicitly check whether the value of ProgressBar.announce and Slider.announce is undefined or null.
This change allows setting the value of ProgressBar.announce to an empty string and still getting an empty string instead of the actual value when the announce getter is called.

## References

LUI-792

## Testing

1. Navigate to the ScrollWrapper component.
2. Turn on the announcer.
3. The announcer should start announcing the text.
4. Slide the slider down.
5. The announcer should not announce the slider value, but continue announcing the text.

Also make sure this changes didn't effect ProgressBar and Slider announcer.

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/ProgressBar/ProgressBar.js
+++ b/packages/@lightningjs/ui-components/src/components/ProgressBar/ProgressBar.js
@@ -100,7 +100,8 @@ export default class ProgressBar extends Base {
   }
 
   get announce() {
-    if (this._announce !== undefined && this._announce !== null) {
+    //use loose equality here because we want to check for both null and undefined values
+    if (this._announce != undefined) {
       return this._announce;
     }
 
--- a/packages/@lightningjs/ui-components/src/components/ProgressBar/ProgressBar.js
+++ b/packages/@lightningjs/ui-components/src/components/ProgressBar/ProgressBar.js
@@ -37,7 +37,7 @@ export default class ProgressBar extends Base {
   }
 
   static get properties() {
-    return ['progress', 'disableAnnouncer'];
+    return ['progress'];
   }
 
   static get tags() {
@@ -47,13 +47,12 @@ export default class ProgressBar extends Base {
   _construct() {
     super._construct && super._construct();
     this._progress = 0;
-    this.disableAnnouncer = false;
   }
 
   _update() {
     this._updateTextures();
     this._updateProgress();
-    if (this._progressChanged && !this.disableAnnouncer) {
+    if (this._progressChanged) {
       this.fireAncestors('$announce', this.announce);
       this._progressChanged = false;
     }
@@ -101,7 +100,7 @@ export default class ProgressBar extends Base {
   }
 
   get announce() {
-    if (this._announce) {
+    if (this._announce != undefined) {
       return this._announce;
     }
 
--- a/packages/@lightningjs/ui-components/src/components/ProgressBar/ProgressBar.js
+++ b/packages/@lightningjs/ui-components/src/components/ProgressBar/ProgressBar.js
@@ -100,7 +100,7 @@ export default class ProgressBar extends Base {
   }
 
   get announce() {
-    if (this._announce != undefined) {
+    if (this._announce !== undefined && this._announce !== null) {
       return this._announce;
     }
 
--- a/packages/@lightningjs/ui-components/src/components/ProgressBar/ProgressBar.js
+++ b/packages/@lightningjs/ui-components/src/components/ProgressBar/ProgressBar.js
@@ -100,8 +100,7 @@ export default class ProgressBar extends Base {
   }
 
   get announce() {
-    //use loose equality here because we want to check for both null and undefined values
-    if (this._announce != undefined) {
+    if (this._announce !== undefined && this._announce !== null) {
       return this._announce;
     }
 
--- a/packages/@lightningjs/ui-components/src/components/ProgressBar/ProgressBar.js
+++ b/packages/@lightningjs/ui-components/src/components/ProgressBar/ProgressBar.js
@@ -37,7 +37,7 @@ export default class ProgressBar extends Base {
   }
 
   static get properties() {
-    return ['progress'];
+    return ['progress', 'disableAnnouncer'];
   }
 
   static get tags() {
@@ -47,12 +47,13 @@ export default class ProgressBar extends Base {
   _construct() {
     super._construct && super._construct();
     this._progress = 0;
+    this.disableAnnouncer = false;
   }
 
   _update() {
     this._updateTextures();
     this._updateProgress();
-    if (this._progressChanged) {
+    if (this._progressChanged && !this.disableAnnouncer) {
       this.fireAncestors('$announce', this.announce);
       this._progressChanged = false;
     }
@@ -110,6 +111,6 @@ export default class ProgressBar extends Base {
     } else if (progress < 0) {
       progress = 0;
     }
-    return `${progress * 100}%`;
+    return `${Math.round(progress * 100)}%`;
   }
 }
--- a/packages/@lightningjs/ui-components/src/components/ScrollWrapper/ScrollSlider.js
+++ b/packages/@lightningjs/ui-components/src/components/ScrollWrapper/ScrollSlider.js
@@ -27,8 +27,4 @@ export default class ScrollSlider extends Slider {
   static get __themeStyle() {
     return styles;
   }
-
-  _init() {
-    this.announce = '';
-  }
 }
--- a/packages/@lightningjs/ui-components/src/components/ScrollWrapper/ScrollSlider.js
+++ b/packages/@lightningjs/ui-components/src/components/ScrollWrapper/ScrollSlider.js
@@ -27,4 +27,8 @@ export default class ScrollSlider extends Slider {
   static get __themeStyle() {
     return styles;
   }
+
+  _init() {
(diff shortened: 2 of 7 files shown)


anthony9187:
were we able to debug why the Slider's value is getting announced even though an override string is being passed to it? we shouldn't need to supply a separate property to override the announcer if that's working properly

THoj13:
> were we able to debug why the Slider's value is getting announced even though an override string is being passed to it? we shouldn't need to supply a separate property to override the announcer if that's working properly

@anthony9187 I was taking a look at this with @soumyaloka, this is what we found:
1. ProgressBar calls `this.fireAncestors('$announce', this.announce)` [each time it's `progress` changes](https://github.com/rdkcentral/Lightning-UI-Components/blob/develop/packages/%40lightningjs/ui-components/src/components/ProgressBar/ProgressBar.js#L56). ProgressBar.announce reads out either the announce set on it, or it's `progress` value (which is happening in this case). 
2. In `withAnnouncer` there's a [listener for that signal that cancels anything in queue to speak and speaks what that signal was called with](https://github.com/rdkcentral/Lightning-UI-Components/blob/develop/packages/%40lightningjs/ui-components/src/mixins/withAnnouncer/index.js#L198-L199).
So it seems like if ProgressBar.progress is changed (or any other component firing an $announce signal), it'll overwrite anything else queued up in announcer.

anthony9187:
> 

It looks like [Slider does this as well](https://github.com/rdkcentral/Lightning-UI-Components/blob/2a4d2765ae1493d84e49109ea1569cd900aa9a9d/packages/%40lightningjs/ui-components/src/components/Slider/Slider.js#L113).

My gut says Slider should be overriding whatever ProgressBar is trying to announce. This would normally happen in our setup but since Slider imports and leverages ProgressBar instead of extending it, we aren't overriding it's announce accessors. A quick fix may be to ensure Slider is passing an empty string to ProgressBar in these situations and let Slider handle announcing entirely. I'm not sure if the empty string will still cause ProgressBar to interrupt whatever is being announced though.

I'm not sure we should add an arbitrary property to disable announcing for this one specific component. If at all, this feature should be built into announcer so we don't wind up with something related to Announcer not being a part of Announcer's API. One of the situations where the ability to intercept a thing would've been useful :/

soumyaloka:
> > 
> 
> It looks like [Slider does this as well](https://github.com/rdkcentral/Lightning-UI-Components/blob/2a4d2765ae1493d84e49109ea1569cd900aa9a9d/packages/%40lightningjs/ui-components/src/components/Slider/Slider.js#L113).
> 
> My gut says Slider should be overriding whatever ProgressBar is trying to announce. This would normally happen in our setup but since Slider imports and leverages ProgressBar instead of extending it, we aren't overriding it's announce accessors. A quick fix may be to ensure Slider is passing an empty string to ProgressBar in these situations and let Slider handle announcing entirely. I'm not sure if the empty string will still cause ProgressBar to interrupt whatever is being announced though.
> 
> I'm not sure we should add an arbitrary property to disable announcing for this one specific component. If at all, this feature should be built into announcer so we don't wind up with something related to Announcer not being a part of Announcer's API. One of the situations where the ability to intercept a thing would've been useful :/

Hey, Sorry I have missed your comments yesterday. Yeah, It may be worth to build this feature into Announcer itself. This would ensure that any component can easily intercept and control what is being announced, rather than relying on each component to pass an empty string.

What do you think @THoj13 ?

THoj13:
Agreed, that's a good point that if we're going to allow disabling announcer on a case-by-case basis that should be in the Announcer API (not in individual components).

I think passing an `announce: ''` from Slider to ProgressBar will still interrupt the current speech text with an empty string. One option I can think of to change that would be to update `withAnnouncer.$announce` to do nothing if it's invoked with an empty string (or any falsey value).

ex. exit the method before it interrupts the current speech an voices out what was passed to it
```js
// withAnnouncer/index.js
$announce(toAnnounce, { append = false, notification = false } = {}) {
    if (!toAnnounce) {
        return;
    }
    ...
}
```

packages/@lightningjs/ui-components/src/components/ProgressBar/ProgressBar.js
```diff
@@ -100,7 +100,7 @@ export default class ProgressBar extends Base {
   }
 
   get announce() {
-    if (this._announce) {
+    if (this._announce != undefined) {
```
anthony9187: can we use `!==` here?

packages/@lightningjs/ui-components/src/components/ScrollWrapper/ScrollSlider.js
```diff
@@ -27,4 +27,8 @@ export default class ScrollSlider extends Slider {
   static get __themeStyle() {
     return styles;
   }
+
+  _init() {
+    this.announce = '';
```
anthony9187: I'm trying to decide if I feel better about this here, or in the ScrollWrapper. It's still overridable since it's in _init, but I'm on the fence. I know this isn't used anywhere else, but just in case

packages/@lightningjs/ui-components/src/components/Slider/Slider.js
```diff
@@ -323,9 +324,10 @@ export default class Slider extends Base {
   }
 
   get announce() {
-    return (
-      this._announce || (this.value !== undefined && this.value.toString())
-    );
+    if (this._announce != undefined) {
```
anthony9187: same here, can we use `!==`?

packages/@lightningjs/ui-components/src/components/ScrollWrapper/ScrollSlider.js
soumyaloka: we made the change in the ProgressBar and Slider components because they both use the _announce property in a similar way.

packages/@lightningjs/ui-components/src/components/ProgressBar/ProgressBar.js
soumyaloka: Tim had a reason here : https://github.com/rdkcentral/Lightning-UI-Components/pull/208#discussion_r1188764673

anthony9187: I don't think I follow - what are we specifically leaning on that requires the use of loose equality instead of strict? the comment explains why we're adding the comparison to undefined, not the use loose equality

anthony9187: @THoj13 ☝️

soumyaloka: Tim's explanation for using explicit null/undefined checks thought it is valid to use the loose quality but I think using the strict will also work.

soumyaloka: Ok, I will change it to strict operator then.

packages/@lightningjs/ui-components/src/components/ScrollWrapper/ScrollSlider.js
anthony9187: I understand that, I'm just thinking about explicitly setting the default here in `_init`. I also feel like we usually try to set defaults in `_construct`, but I'll defer to @erautenberg on that

erautenberg: Just caught up. I think moving this from init to construct makes sense, and maybe we can do `this._announce = '';` instead.

packages/@lightningjs/ui-components/src/components/ProgressBar/ProgressBar.js
THoj13: @anthony9187 I used loose equality so that `this._announce` being `null` or `undefined` would have the same falsey result since:
```js
// this._announce != undefined
null != undefined // false
undefined != undefined // false

// this._announce !== undefined
null !== undefined // true
undefined !== undefined // false
```

I'm not sure if `null` would ever be the value of `this._announce` though, so perhaps doesn't need to be accounted for.

anthony9187: Thanks for the explanation. I'm ok with using loose equality here then, let's just make sure we comment both of the spots(cause I'll definitely just go "hm let me fix this") and add some unit tests around it

packages/@lightningjs/ui-components/src/components/Slider/Slider.js
```diff
@@ -323,9 +324,10 @@ export default class Slider extends Base {
   }
 
   get announce() {
-    return (
-      this._announce || (this.value !== undefined && this.value.toString())
-    );
+    if (this._announce !== undefined && this._announce !== null) {
+      return this._announce;
+    }
+    return this.value !== undefined && this.value.toString();
```
THoj13: Since the if statement before this line already checks `this._announce !== undefined`, that check can be removed from the check in this last return statement.
```suggestion
    return this.value.toString();
```

THoj13: Since the if statement before this line already checks `this._announce !== undefined`, that check can be removed from the check in this last return statement.
```suggestion
    return this.value.toString();
```

anthony9187 (approved):

ImCoolNowRight (approved):

PR #206, opened 2023-05-08
chore: add title compliance check

soumyaloka:
## References

In this PR, I have added a check to the PR workflow which will check the PR title to make sure it matches our implementation of the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0-beta.2/) syntax. 

## Testing

LUI-759

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -4,8 +4,7 @@ on:
   pull_request:
     types: [opened, edited, reopened, synchronize, ready_for_review]
 
-permissions:
-  pull-requests: read-all
+permissions: read-all
 
 jobs:
   quality:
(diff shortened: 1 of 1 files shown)


ImCoolNowRight (approved):

arwehrman (approved):

.github/workflows/pull-request.yml
```diff
@@ -1,8 +1,28 @@
...
   quality:
     name: Quality
     uses: ./.github/workflows/quality.yml
+
+  check-pr-title:
+    name: Check PR title
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
```
ImCoolNowRight: Can you try moving this checkout step to the Quality job, before line 7? I think we need this there to "pull in" our repo before running that Quality job.

soumyaloka: I can move that to quality job but not before permissions I believe

PR #208, opened 2023-05-09
fix(withAnnouncer): ignore empty speech strings

THoj13:
## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/ProgressBar/ProgressBar.js
+++ b/packages/@lightningjs/ui-components/src/components/ProgressBar/ProgressBar.js
@@ -37,7 +37,7 @@ export default class ProgressBar extends Base {
   }
 
   static get properties() {
-    return ['progress', 'disableAnnouncer'];
+    return ['progress'];
   }
 
   static get tags() {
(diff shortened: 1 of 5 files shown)


packages/@lightningjs/ui-components/src/components/ProgressBar/ProgressBar.js
```diff
@@ -101,7 +100,7 @@ export default class ProgressBar extends Base {
   }
 
   get announce() {
-    if (this._announce) {
+    if (this._announce != undefined) {
```
THoj13, line 103: Updated the conditional from evaluating truthy/falsey of `this._announce` to explicitly check whether or not it is `undefined` or `null`. This allows setting `ProgressBar.announce` to `''` and still getting that empty string instead of its value when the announce getter is called.

The same change is made in `Slider`.

soumyaloka: Thanks for the update, I think this is the better way to handle the empty string scenario in both ProgressBar and Slider.

soumyaloka (approved):

PR #210, opened 2023-05-12
fix(Provider): add setVisibleCount method for out of range values

soumyaloka:
## Description

This PR adds a set method to the _setVisibleCount method which make sure that negative values and values outside of the bounds are not accepted

## References

LUI-731

## Testing

1. Navigate to the Provider and MetadataCardComponent.
2. Try setting the visible count value to a negative number.
3. Confirm that the value is automatically set to 1.
4. Observe that a warning message is logged to the console, stating that the visible count must be greater than or equal to 1 and vice versa for the numbers that exceeds providers length.
5. Try setting the visible count value to a number within the range of 1 to the number of providers length and confirm that the value is set correctly without any warning message being logged to the console.


erautenberg (approved):

arwehrman (approved):

ImCoolNowRight (changes requested):
Looks great, thank you!

I tried the below code as a way to get a `console.warn` working, and I think it does the job.
```js
_setVisibleCount(count) {
  if (count < 1 || count > this.providers.length) {
    console.warn(
      "'visibleCount' must be greater than or equal to 1 and less than or equal to the number of providers."
    );
  }
  return count;
}
```

soumyaloka:
> Looks great, thank you!
> 
> I tried the below code as a way to get a `console.warn` working, and I think it does the job.
> 
> ```js
> _setVisibleCount(count) {
>   if (count < 1 || count > this.providers.length) {
>     console.warn(
>       "'visibleCount' must be greater than or equal to 1 and less than or equal to the number of providers."
>     );
>   }
>   return count;
> }
> ```

Ok, let me try that. Thanks!

ImCoolNowRight (approved):

PR #211, opened 2023-05-15
fix(Tile): fixes transitions on metadata metadata

arwehrman:
## Description

This PR does the following:

- fixes metadataPatch so bottom animation works  every time Tile goes from unfocus to focus
- restructures the code, breaking down components into sections for readability
- adds more logic to the shouldShowMetadata to account for all possible scenarios instead of just persistentMetadata and isFocused

## References

LUI-658

## Testing

- pull down branch
- `yarn start`
-  go to Tile story
- select `inset` in controls
- click `focus` in controls, bottom metadata data should move from bottom to top
- click `unfocus` and `focus` again the transition should perform as expected

## Checklist


packages/@lightningjs/ui-components/src/components/Tile/Tile.js
```diff
@@ -250,26 +176,24 @@ export default class Tile extends Surface {
...
     this._Background.alpha = 0; // Since the image is loaded the surface does not need to be shown
   }
 
+  /* ------------------------------ Badge ------------------------------ */
   _updateBadge() {
-    // Remove Badge if no longer required
-    if (!this.badge?.title || this._isCircleLayout) {
-      if (this._Badge) {
-        this._Content.patch({
-          Badge: undefined
```
erautenberg: Why was this removed? I noticed that the Label equivalent is still there

packages/@lightningjs/ui-components/src/components/Tile/Tile.js
```diff
@@ -326,14 +252,37 @@ export default class Tile extends Surface {
...
+  get _badgeLabelTransitions() {
+    return {
+      y: [
+        this._shouldShowBadgeLabel ? this.style.paddingY : 0,
+        this.__shouldShowBadgeLabel
+          ? this.style.animationEntrance
+          : this.style.animationExit
+      ],
+      alpha: [
+        this._shouldShowBadgeLabel && !this._isCircleLayout ? 1 : 0.001,
```
erautenberg: I'm a little confused by all the alpha logic for Badge and Label. You have this here, but also basically the same thing in the patch for the updateBadge and updateLabel methods.

packages/@lightningjs/ui-components/src/components/Tile/Tile.js
```diff
@@ -394,7 +347,7 @@ export default class Tile extends Surface {
       }
       return;
     }
-
+    // --------------------------------------//
```
erautenberg: not sure what this bar is adding or signifying?

packages/@lightningjs/ui-components/src/components/Tile/Tile.js
```diff
@@ -441,11 +394,49 @@ export default class Tile extends Surface {
...
+  /* ------------------------------ Metadata  ------------------------------ */
+
+  get _shouldShowMetadata() {
+    return this._persistentMetadata || this._isFocusedMode;
+  }
+
+  get _isInsetMetadata() {
+    return this._metadataLocation === 'inset';
+  }
+  // called in animateMetadata
```
erautenberg: can this comment get moved inside the method or something? At a minimum, let's add an empty line after line 409.

erautenberg (changes requested):
The Badge seems to be broken
Screen Shot 2023-05-16 at 10 11 19 AM

Also, let's check with Sean on the animation in and out. I think when going from focus to unfocus, the inset is supposed to slide back down (opposite of the slide up it does on the way in). Also it doesn't seem like it is alpha-ing in correctly, it appears at the bottom, then slides up, but doesn't fade in as it slides up.

packages/@lightningjs/ui-components/src/components/Tile/Tile.js
arwehrman: left over comment from when I was troubleshooting something. Will remove

packages/@lightningjs/ui-components/src/components/Tile/Tile.js
arwehrman: I'll remove the comment. It's a leftover from my notes

packages/@lightningjs/ui-components/src/components/Tile/Tile.js
arwehrman: I removed it when I was figuring out whether we patch the component in as `undefined` or alpha out the component

arwehrman: I can add it back

erautenberg (approved):

ImCoolNowRight (approved):

PR #212, opened 2023-05-15
fix: make sure context listeners are removed on _detach

joshhowenstine:
## Description

Cleanup themeUpdate event listeners when when a component in destroyed in withThemeStyles as well as withExtensions mixins.

This PR also will convert every base64 encoded image stored in a theme to a URL using `URL.createObjectURL`.

It has been observed that base64 encoded strings seamed to get cached in memory causing performance issues. By converting these images into urls we hope to alleviate this issue. 

## References

- [LUI-817](https://ccp.sys.comcast.net/browse/LUI-817)
- [LUI-825](https://ccp.sys.comcast.net/browse/LUI-825)

## Testing

Create and destroy an Artwork component or Tile many times. Memory should be freed up once the component is removed. 

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/globals/context/theme-manager.js
+++ b/packages/@lightningjs/ui-components/src/globals/context/theme-manager.js
@@ -274,7 +274,7 @@ class ThemeManager {
   _clearCache() {
     // Clean up any base64 image that were converted to blobs using createObjectURL
     base64Cache.filter(image => {
-      URL.revokeObjectURL(image);
+      if (window.URL && typeof window.URL.revokeObjectURL === 'function') URL.revokeObjectURL(image);
       return false;
     });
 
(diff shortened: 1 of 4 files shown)


erautenberg (approved):
Makes sense code-wise, but will defer to Carolyn and the rest of her team to double-check in their app.

anthony9187 (approved):

packages/@lightningjs/ui-components/src/globals/context/theme-manager.js
```diff
@@ -37,7 +37,79 @@ const merge = {
...
+      const byteArray = new Uint8Array(byteNumbers);
+      byteArrays.push(byteArray);
+    }
+
+    const blob = new Blob(byteArrays, { type: mimeType });
+    const blobURL = URL.createObjectURL(blob);
+
+    return blobURL;
+  } catch (error) {
+    logger.info('Unable to convert base64 image to URL');
```
anthony9187, line 108: do we want to include the stack trace here?

chiefcll (approved):

packages/@lightningjs/ui-components/src/globals/context/theme-manager.js
joshhowenstine: Good point, we probably should. Would it be go to add a follow up ticket to add stack trace support to all of our logs?

anthony9187: yeah i think that works

PR #213, opened 2023-05-16
fix(CardContentVerticalSmall): remove unused control

soumyaloka:
## Description

This PR removed the metadata details property from cardContentArgs and cardContentArgsTypes and also shouldCollapse control has been removed from the CardContent variant components because this control is already included in the parent component (CardContent) and since all other variants are extending it, there is no need for a duplicate control.

## References

LUI-818

Testing:

1. Navigate to the CardContentVerticalSmall component in Storybook.
2. Verify that the delete control is not present in the component's story.
3. Also make sure shouldCollapse property is working as expected in each variant of the CardContent component.

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/CardContent/CardContentVerticalSmall.stories.js
+++ b/packages/@lightningjs/ui-components/src/components/CardContent/CardContentVerticalSmall.stories.js
@@ -24,7 +24,6 @@ import { createModeControl, generateSubStory } from '../../docs/utils';
 import CardContentVerticalSmallComponent from './CardContentVerticalSmall';
 import { CardContent } from './CardContent.stories';
 import { CATEGORIES } from '../../docs/constants';
-import { controlDescriptions } from '../../docs/constants';
 
 export default {
   title: `${CATEGORIES[128]}/CardContentVerticalSmall`,
@@ -49,22 +48,22 @@ export const CardContentVerticalSmall = args =>
   };
 
 CardContentVerticalSmall.storyName = 'CardContentVerticalSmall';
+const cardContentArgs = CardContent.args;
+delete cardContentArgs.metadata_details;
 
 CardContentVerticalSmall.args = {
   mode: 'focused',
   ...CardContent.args
 };
+
+const cardContentArgsTypes = CardContent.argTypes;
+delete cardContentArgsTypes.metadata_details;
+
 CardContentVerticalSmall.argTypes = {
   ...createModeControl({ summaryValue: CardContentVerticalSmall.args.mode }),
-  shouldCollapse: {
-    control: 'boolean',
-    description: controlDescriptions.shouldCollapse,
-    table: {
-      defaultValue: { summary: false }
-    }
-  },
   ...CardContent.argTypes
 };
+
 CardContentVerticalSmall.parameters = {
   argActions: CardContent.tileProps.argActions('CardContentVerticalSmall')
 };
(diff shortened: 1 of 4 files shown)


packages/@lightningjs/ui-components/src/components/CardContent/CardContentVerticalSmall.stories.js
```diff
@@ -49,22 +48,22 @@ export const CardContentVerticalSmall = args =>
   };
 
 CardContentVerticalSmall.storyName = 'CardContentVerticalSmall';
+const cardContentArgs = CardContent.args;
+delete cardContentArgs.metadata_details;
```
erautenberg, line 54: everything seems fine, but does it matter that this technically mutates the object? Maybe line 51 should be an Object.assign or JSON.parse(JSON.stringify?

soumyaloka: Yeah, How about adding spread operator (...) to create shallow copies
```
const cardContentArgs = { ...CardContent.args };
delete cardContentArgs.metadata_details;
```

ImCoolNowRight:
I'm still seeing the Details section under Metadata in the stroy.
Screen Shot 2023-05-16 at 4 10 40 PM

soumyaloka:
> I'm still seeing the Details section under Metadata in the stroy. Screen Shot 2023-05-16 at 4 10 40 PM

It looks like it broke after latest commit.

packages/@lightningjs/ui-components/src/components/CardContent/CardContentVerticalSmall.stories.js
```diff
@@ -50,21 +49,24 @@ export const CardContentVerticalSmall = args =>
 
 CardContentVerticalSmall.storyName = 'CardContentVerticalSmall';
 
+//Creating a shallow copy of CardContent.args object and removing metadata details property from cardContentArgs
+const cardContentArgs = { ...CardContent.args };
+delete cardContentArgs.metadata_details;
+
 CardContentVerticalSmall.args = {
   mode: 'focused',
   ...CardContent.args
```
ImCoolNowRight: I believe this should be `cardContentArgs` instead of `CardContent.args`?

packages/@lightningjs/ui-components/src/components/CardContent/CardContentVerticalSmall.stories.js
```diff
@@ -50,21 +49,24 @@ export const CardContentVerticalSmall = args =>
...
 CardContentVerticalSmall.argTypes = {
   ...createModeControl({ summaryValue: CardContentVerticalSmall.args.mode }),
-  shouldCollapse: {
-    control: 'boolean',
-    description: controlDescriptions.shouldCollapse,
-    table: {
-      defaultValue: { summary: false }
-    }
-  },
   ...CardContent.argTypes
```
ImCoolNowRight: Same as above? It looks like you're creating copies based off of `CardContent` but not using them?

packages/@lightningjs/ui-components/src/components/CardContent/CardContentVerticalSmall.stories.js
soumyaloka: Oh yeah, you are right. I didn't assign the new copy.

ImCoolNowRight (approved):

THoj13 (approved):

PR #214, opened 2023-05-16
fix(Tab): fix icon rendering in Tab component

soumyaloka:
## Description

In this PR, I have updated to use the icon property instead of src for specifying the icon and also wrapped content color within the style object.

## References

LUI-822

## Testing

1. Open the Tab and set Icon in icon control.
2. Click on the "Base" theme.
3. Confirm that the icon remains visible at all times, even when the title is removed or the mode is changed. The icon should persist until it is explicitly set to null.

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/TabBar/Tab.js
+++ b/packages/@lightningjs/ui-components/src/components/TabBar/Tab.js
@@ -134,24 +134,20 @@ export default class Tab extends Surface {
   }
 
   _updateContent() {
-    const iconWidth = this._iconW || 0;
-    const titleWidth = this.title
-      ? this.style.iconMarginRight + this._textW
-      : 0;
     this._Content.patch({
-      w: iconWidth + titleWidth,
+      w:
+        this._iconW +
+        (this.title ? this.style.iconMarginRight : 0) +
+        this._textW,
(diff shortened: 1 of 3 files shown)


packages/@lightningjs/ui-components/src/components/TabBar/Tab.js
```diff
@@ -116,12 +123,13 @@ export default class Tab extends Surface {
       y: this._Content.h / 2
     };
     if (this.icon) {
-      textPatch.mountX = 0;
       textPatch.x = this._iconW + this.style.iconMarginRight;
+      textPatch.mountX = 0;
     } else {
-      textPatch.mountX = 0.5;
       textPatch.x = this._Content.w / 2;
+      textPatch.mountX = 0.5;
```
erautenberg, line 130: Does this ordering actually matter? I'm a little surprising we wouldn't want it to be the original order, so just curious

soumyaloka: Sorry, I just updated order for the sake of code organization while debugging the issue.

ImCoolNowRight (approved):
Looks good! I'm assuming the non-Tab files are just due to needing a rebase.

anthony9187 (approved):

PR #222, opened 2023-05-22
feat(utils): add function to attach onAfterUpdate function that runs sideEffect on specific prop changes

THoj13:
## Description
This is a follow up PR to https://github.com/rdkcentral/Lightning-UI-Components/pull/200 to abstract the `onAfterUpdate` logic used in `NavigationManager._withAfterUpdate` to a utility function that can be reused in other components.

Adds utility function, `watchForUpdates`, which adds an `onAfterUpdate` function to a component. The function will be invoked when any of the supplied `watchProps` change.

## References
LUI-797

## Testing
Compare stories in the following components to those on the deployed storybook instance. No visible changes between the branches should be present: 
- Row 
- Column
- Button
- Control (subclass of button)
- ControlRow (utilizes both NavigationManager and Control)

## Automation
n/a

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/utils/index.js
+++ b/packages/@lightningjs/ui-components/src/utils/index.js
@@ -17,6 +17,7 @@
  */
 
 import lng from '@lightningjs/core';
+import logger from '../globals/context/logger';
 
 /**
  *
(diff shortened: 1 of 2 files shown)


packages/@lightningjs/ui-components/src/utils/index.js
```diff
@@ -743,6 +743,51 @@ export function createConditionalZContext(component, zOffset) {
...
+/**
+ * Runs a side effect function when any values of specified properties on a component change.
+ * @param {Object} options - defines the `component`, `watchProps`, and `sideEffect` options
+ * @param {lng.Component} options.component - Lightning component on which property changes will watched
+ * @param {String[]} options.watchProps - properties that when their value changes a side effect function is invoked
+ * @param {Function} options.sideEffect - function to be invoked when a watched property's value changes
+ * @returns {lng.Component}
+ */
+export function watchForUpdates({
+  component = lng.Component,
```
anthony9187: do we want to have a default here? seems like we'd be adding the onAfterUpdate methods to an arbitrary component if one isn't provided as a parameter. I might be misunderstanding the function

THoj13: Nope, that's correct. Maybe throwing an error would be more helpful here so that it's clear that an important parameter wasn't supplied.

anthony9187: yeah if the component parameter is required for this to work we should either let the function break or throw some kind of error

erautenberg (approved):

anthony9187 (approved):

PR #227, opened 2023-05-25
feat(CardRadio): reupload - refactoring and apply theming

jaylen-creative-engineer:
## Description

Adding the CardRadio Component to the components library.

This is a reupload due to challenges with the SSH Key Signing. 

Huge thanks to @soumyaloka for her help.

I will aim to respond to feedback quickly.

## References

LUI-500

## Checklist


packages/@lightningjs/ui-components/src/components/CardRadio/CardRadio.mdx
```diff
@@ -0,0 +1,73 @@
...
+| name     | type   | readonly | description                                                   |
+| -------- | ------ | -------- | ------------------------------------------------------------- |
+| subtitle | string | false    | The text to be displayed in the subtitle section of the card. |
+
+### Style Properties
+
+| name                 | type            | description                                      |
+| -------------------- | --------------- | ------------------------------------------------ |
+| alpha                | number          | transparency applied to text and radio component |
+| descriptionTextStyle | string \|object | styles applied to the description text.          |
```
ImCoolNowRight: Can you either remove the space before the `|` between the types or add a space after? Same with line 73.

packages/@lightningjs/ui-components/src/components/CardRadio/CardRadio.js
```diff
@@ -0,0 +1,125 @@
...
+          ...this.style.subtitleTextStyle,
+          wordWrapWidth: this._calculateTextWidth()
+        }
+      }
+    });
+  }
+
+  _updateRadio() {
+    const { iconWidth, iconHeight } = this.style;
+    const RadioPatch = {
```
ImCoolNowRight: Can you make this `radioPatch` instead? Even with patch objects, we try to stick to camelCase.

ImCoolNowRight (changes requested):

ImCoolNowRight (approved):

packages/@lightningjs/ui-components/src/components/CardRadio/CardRadio.js
```diff
@@ -0,0 +1,125 @@
...
+    this._updateRadio();
+    this._updateDescriptionPosition();
+    this._updateSubtitlePosition();
+  }
+
+  _updateTitle() {
+    super._updateTitle();
+    this._Title.patch({
+      signals: {
+        textBoxChanged: '_update'
```
erautenberg: Is this necessary? The Title textbox already has this set up in the Card template

```js
      Title: {
        type: TextBox,
        signals: {
          textBoxChanged: 'queueRequestUpdate'
        }
      }
```

packages/@lightningjs/ui-components/src/components/CardRadio/CardRadio.js
```diff
@@ -0,0 +1,125 @@
...
+      y:
+        (this._Title.style.textStyle.lineHeight - iconHeight) / 2 +
+        this.style.paddingVertical +
+        this.style.paddingVertical / 4,
+      checked: true
+    };
+    if (!this._Radio) {
+      radioPatch.type = Radio;
+    }
+    this.patch({ Radio: radioPatch, alpha: this.style.alpha });
```
erautenberg: I think this alpha needs to be moved to the Radio's style object since the Radio component has this method
```js
  _updateOpacity() {
    this.applySmooth(this, { alpha: this.style.alpha });
  }
```

erautenberg (changes requested):
Need to export CardRadio in src/components/index.js and index.d.ts

erautenberg:
Could you add a control for the Radio to the story using the "generateSubStory" method? That way we can test toggling the radio on and off.

packages/@lightningjs/ui-components/src/components/Card/Card.js
```diff
@@ -43,7 +43,7 @@ export default class Card extends Surface {
       Title: {
         type: TextBox,
         signals: {
-          textBoxChanged: 'queueRequestUpdate'
+          textBoxChanged: '_update'
```
erautenberg: Why was this changed to update? The queueRequestUpdate makes sure things get funneled through the update manager and we are batching and debouncing appropriately.

packages/@lightningjs/ui-components/src/components/Card/CardRadio.d.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export type CardRadioStyle = CardTitleStyle & {
+  descriptionTextStyle: TextBoxStyle;
+  iconHeight: number;
+  iconWidth: number;
+  subtitleTextStyle: TextBoxStyle;
+};
+export default class CardRadio extends CardTitle {
+  radio?: Record<string, unknown>;
+  subtitle?: string;
```
erautenberg: should this type actually match the Textbox content? (string or array)

packages/@lightningjs/ui-components/src/components/Card/CardRadio.js
```diff
@@ -0,0 +1,118 @@
...
+    });
+  }
+
+  _updateRadio() {
+    const { iconWidth, iconHeight } = this.style;
+    const radioPatch = {
+      ...this.radio,
+      mode: this.mode,
+      w: iconWidth,
+      h: iconHeight,
```
erautenberg: `icon` is a misleading term here. Unless the designers specified that the radio should be a larger size on the CardRadio component than the normal Radio component, then I don't think this is necessary anyway since the user can pass in something like this
```js
CardRadio: {
  radio: {
    style: { w: 40, h: 40 }
  }
}
```

If the height DOES need to be increased per design specs, then let's just add a nested radio object in styles so we would have something like this in this updateRadio method:
```js
  const radioPatch = {
    ...this.styles.radio,
    ...this.radio,
    mountX: 1
    x: this.w - this.style.paddingHorizontal,
    y: this.style.paddingVertical
  };
```
(that Y position is also overly complicated, it should just be the padding value)

erautenberg: And honestly, if the size does need to change for Xfinity design specs, we should just do that in the Xfinity theme and I would remove that altogether here, so you'd only have:
```js
  const radioPatch = {
    ...this.radio,
    mountX: 1
    x: this.w - this.style.paddingHorizontal,
    y: this.style.paddingVertical
  };
```

packages/@lightningjs/ui-components/src/components/Card/CardRadio.d.ts
soumyaloka: so something like this?
 subtitle?: string | TextContent[];

anthony9187: yep, since subtitle is getting passed straight through to TextBox.content, it should match the type. so `string | TextContent[]`

soumyaloka: Thank You @anthony9187

erautenberg (approved):

PR #229, opened 2023-05-26
feat(InlineContent): add support for max lines and truncation

THoj13:
## Description
Both `maxLines` and `maxLinesSuffix` have been supported in the `TextBox` component, as long as just a text string is passed to the component. Prior to this change, neither of those properties were supported when using inline content strings or objects (typically representative of a combination of strings, badges and icons rendered together).

- InlineContent
  - support for `maxLines` property
  - support for `maxLinesSuffix` property
  - both of these should have the same effect as when used with TextBox (which renders a Lightning Text texture, see more details in the [Lightning Text texture documentation](https://lightningjs.io/docs/#/lightning-core-reference/RenderEngine/Textures/Text?id=properties))
- TextBox
  - supports applying `maxLines` and `maxLinesSuffix` to inline content strings or objects via `style.textStyle` 

## References
LUI-810

## Testing
Two stories were added to demonstrate this new functionality:
- TextBox: With Inline Content Truncation
- InlineContent: With Truncation

## Automation
N/A

## Checklist


erautenberg (approved):
This looks so good, great job! Would you mind just adding the maxLines control to the story so we can test with other variations on the fly?

ImCoolNowRight (approved):
Awesome job! +1 for Emily's request for the extra story control, but otherwise looks great!

THoj13:
After adding `maxLines` and `maxLines` suffix as args on all the InlineContent stories, I'm seeing issues on the following stories:
- With Parsing: if `maxLines` is set, the component is no longer displayed
- With Truncation: setting `maxLines` to 1 is wrapping the last word onto a second line instead of truncating the content to 1 line

Adding a DO NOT MERGE label until these are resolved.

THoj13:
> After adding maxLines and maxLines suffix as args on all the InlineContent stories, I'm seeing issues on the following stories:
>
>  - With Parsing: if maxLines is set, the component is no longer displayed
>  - With Truncation: setting maxLines to 1 is wrapping the last word onto a second line instead of truncating the content to 1 line 

These have been resolved.

PR #230, opened 2023-05-30
fix: disable image worker for blob urls in Icon component

joshhowenstine:
## Description
In the previous release, base64 images from a Lightning UI Component theme were converted into a BlobURL to reduce memory usage. However, if the `useImageWorker` option is set to true in the application, these icons fail to load.

This pull request addresses the issue by creating a custom texture for blob images, bypassing the web worker feature only for blob icons.

## References
- [Jira Ticket LUI-857](https://ccp.sys.comcast.net/browse/LUI-857)

## Testing
- In `packages/apps/lightning-ui-docs/index.js`, change the value of `useImageWorker` to true.
- Run Storybook and observe the ScrollWrapper story.
- The arrows should be visible when using image workers.

## Checklist
- [x] Removed all commented code
- [x] Resolved any new console issues
- [x] Ran code linter and formatter
- [x] Ensured test coverage meets repository requirements


erautenberg (changes requested):
Need to roll back the base theme and remove that Artwork fallback, that is Xfinity specific and doesn't belong in the Base theme.

erautenberg (changes requested):
There is something weird going on with the Tile on the Tokens overview story with this change
Screen Shot 2023-06-28 at 5 44 02 PM

ImCoolNowRight:
> There is something weird going on with the Tile on the Tokens overview story with this change Screen Shot 2023-06-28 at 5 44 02 PM

I'm not sure it's due to this PR. I'm seeing the same thing in `develop`.
Screen Shot 2023-06-29 at 2 44 22 PM

ImCoolNowRight:
> There is something weird going on with the Tile on the Tokens overview story with this change Screen Shot 2023-06-28 at 5 44 02 PM

I did some messing around to try to help get this fixed so we can merge it in, and it looks like this is specifically an issue with PNGs. JPGs, both local and remote, have no problem loading in and displaying properly. The Tile story points to a remote JPG, the Theming Foundations is using a local PNG. The only thing I see specific in Lightning Core for how PNGs are handled is a `hasAlpha` property in `@lightningjs/core/src/platforms/browser/WebPlatform.mjs` in the `loadSrcTexture` function, but that's obviously not it.

ImCoolNowRight (approved):

erautenberg (approved):
Doesn't seem to be a side-effect of this PR, but the `shouldScale` prop on Artwork does not seem to be doing anything on the Base or Xfinity theme (checked on IS LUI). Let's get a ticket to track.

PR #238, opened 2023-06-06
fix(Column): update TS structure

jazzyw:
## Description

Updates Column's TS declaration file to match the new TS structure ([@lightningjs/core subclassable components guide](https://github.com/rdkcentral/Lightning/blob/2e4c829be45614290405c222bf102022a6219153/docs/TypeScript/Components/SubclassableComponents.md))

Additional Updates

- Moves the `TransitionObject` to the lui.d.ts file since the object is used in multiple components
- Adds additional properties and accessors to `FocusManager`

## References

LUI-836

## Testing

- Create a new component using `yarn createComponent @lightningjs/ui-components MyComponent`
- Make sure the component extends `Column` (rather than Base)
- Ensure the `Column` properties/methods show up when using the dot syntax or through patch

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Base/Base.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/Base/Base.d.ts
@@ -63,6 +63,8 @@ declare class Base<
    */
   loaded?: Promise<void>;
 
+  // TODO: flag for discussion
+  // internal stuff can be omitted --> `shouldSmooth` is internal?
   shouldSmooth?: boolean;
 
   /**
@@ -100,6 +102,7 @@ declare class Base<
    */
   isFullyOnScreen(): boolean;
 
+  // TODO: for future reference these accessors should technically be public
   /**
    * check if the component is disabled
    */
--- a/packages/@lightningjs/ui-components/src/components/Column/Column.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/Column/Column.d.ts
@@ -20,7 +20,6 @@ import lng from '@lightningjs/core';
 import NavigationManager from '../NavigationManager';
 
 declare namespace Column {
-  // errored until NavigationManager TS updates are merged
   export interface TemplateSpec extends NavigationManager.TemplateSpec {
     /**
      * When navigation between multiple Columns,
@@ -48,16 +47,12 @@ declare class Column<
    */
   checkSkipPlinko(prev: lng.Component, next: lng.Component): lng.Component;
 
-  // TODO: flag for discussion
-  // in the docs, there is a parameter for this function but in Column.js the function has no params
   /**
    * A callback that can be overridden to do something with the items that are currently on screen.
    * This will be called on every new render.
    */
   onScreenEffect(): void;
 
-  // TODO: flag for conversation/discussion
-  // should we include these methods?
   /**
    * Removes the passed in item from the items array and updates the selectedIndex, if necessary
    * @param item component to be removed
--- a/packages/@lightningjs/ui-components/src/components/Column/Column.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/Column/Column.d.ts
@@ -17,26 +17,7 @@
  */
 
 import lng from '@lightningjs/core';
-import NavigationManager, {
-  NavigationManagerStyle
-} from '../NavigationManager/NavigationManager';
-import type { StylePartial } from '../../types/lui';
-
-// TODO: ask --this is exported in NavigationManager; is this even necessary?
-export type TransitionObject = {
-  delay: number;
-  duration: number;
-  timingFunction: string;
-};
-
-// TODO: ask --all these style props are part of NavigationManager; do we need to define the props again? Or should the object just be empty?
-export type ColumnStyle = NavigationManagerStyle & {
-  itemSpacing: number;
-  scrollIndex: number;
-  alwaysScroll: boolean;
-  neverScroll: boolean;
-  itemTransition: TransitionObject;
-};
+import NavigationManager from '../NavigationManager';
 
 declare namespace Column {
   // errored until NavigationManager TS updates are merged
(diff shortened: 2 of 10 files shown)


packages/@lightningjs/ui-components/src/components/Column/Column.d.ts
```diff
@@ -18,32 +18,56 @@
...
+  // Methods
+  /**
+   * Returns the item right before the item that has skipPlinko or before prev if no item has skipPlinko
+   * @param prev component existing right before the current component
+   * @param next component existing right after the current component
+   */
   checkSkipPlinko(prev: lng.Component, next: lng.Component): lng.Component;
+
+  // TODO: flag for discussion
+  // in the docs, there is a parameter for this function but in Column.js the function has no params
```
anthony9187: my assumption would be that this method gets fully overridden so there's no need for any parameters. is it just an empty function in Column.js?

packages/@lightningjs/ui-components/src/components/Base/Base.d.ts
```diff
@@ -63,6 +63,8 @@ declare class Base<
    */
   loaded?: Promise<void>;
 
+  // TODO: flag for discussion
+  // internal stuff can be omitted --> `shouldSmooth` is internal?
```
anthony9187, line 67: @erautenberg are we treating shouldSmooth as internal or should we be documenting this in the d.ts files?

erautenberg: That's a good callout 🤔 I think since it's not underscored, it might be worth adding here. I think we had to use it once or twice in Flex for some Row/Column updates.

packages/@lightningjs/ui-components/src/components/Column/Column.d.ts
jazzyw: Yes in Column.js it's just an empty function. Then does the doc need to remove the param?

packages/@lightningjs/ui-components/src/components/Column/Column.d.ts
```diff
@@ -18,32 +18,56 @@
...
+   */
   onScreenEffect(): void;
 
+  // TODO: flag for conversation/discussion
+  // should we include these methods?
+  /**
+   * Removes the passed in item from the items array and updates the selectedIndex, if necessary
+   * @param item component to be removed
+   */
   $removeItem(item: lng.Component): void;
```
jazzyw, line 60: @erautenberg spoke with Anthony on this but we wanted to confirm with you if we should include these event methods in the TS files?

packages/@lightningjs/ui-components/src/types/lui.d.ts
```diff
@@ -34,3 +34,11 @@ export type StylePartial<T> = {
     : // If it is NOT an object, then just spit back the original value type.
       T[P];
 };
+
+// TODO: flag for discussion
+// more specific naming of this object
+export type TransitionObject = {
```
jazzyw: Leaving a comment here for more visibility. Any thoughts on what we should be naming this object? Anthony first suggested `TransitionSetting`.

packages/@lightningjs/ui-components/src/components/Column/Column.d.ts
anthony9187: yes, lets remove it from the docs

packages/@lightningjs/ui-components/src/components/Base/Base.d.ts
anthony9187: so my next question is, should `shouldSmooth` get added to `Base` since technically it can be set on any component?

packages/@lightningjs/ui-components/src/components/Column/Column.d.ts
jazzyw: Sounds good thank you! Will make the update

packages/@lightningjs/ui-components/src/types/lui.d.ts
ImCoolNowRight: Maybe `TransitionConfig`? I don't mind `TransitionSetting`, but it does feel weird with it having multiple properties and not being plural, which I know we're against.

packages/@lightningjs/ui-components/src/components/Base/Base.d.ts
ImCoolNowRight: Did we ever discuss this? I think it makes sense to add to Base.

packages/@lightningjs/ui-components/src/components/Column/Column.d.ts
jazzyw: From standup this morning, confirmed these should be included if it's useful for the user to access.

packages/@lightningjs/ui-components/src/types/lui.d.ts
jazzyw: From standup, ask Frank if there's already an object that has the transition properties and if so we could use that. From Anthony, if the object has more properties than we need, we can create a subprop that picks only select properties rather than the entire object.

jazzyw: Looks like LightningCore already has something called `TransitionSettings`!

packages/@lightningjs/ui-components/src/components/Base/Base.d.ts
anthony9187: i double checked and this exists in Base, however only in the class body and without any actual documentation. Opened a PR to update those docs https://github.com/rdkcentral/Lightning-UI-Components/pull/266

anthony9187: this comment is literally on Base.d.ts 🤦

anthony9187 (approved):

erautenberg (approved):

PR #240, opened 2023-06-06
fix(Tile): add style prop for spacing between Metadata and ProgressBar

erautenberg:
## Description

The spacing between the Metadata and ProgressBar on the Tile should have a value independent of the existing paddingY and paddingYProgress.

## Testing

Observe that setting Progress to 0 will move the metadata down to the same paddingY value as the badge/label after the ProgressBar is removed, and that the spacing between ProgressBar and Metadata is a different value when present.

## Checklist


soumyaloka:
The spacing in the PR looks good. I have noticed an issue that is not rooted in this PR. It seems that the metadata is moving every time I change the progress which is not expected I guess. I have also observed this issue on the develop branch.

ImCoolNowRight (approved):

ImCoolNowRight (commented):
I did just notice that the metadata is closer to the progress bar in this branch compared to develop. Is that intentional?

packages/@lightningjs/ui-components/src/components/Tile/Tile.js
```diff
@@ -350,14 +350,16 @@ export default class Tile extends Surface {
...
+  get _progressBarY() {
     return (
       (this._ProgressBar &&
-        this._ProgressBar._getTransition('h')._targetValue +
-          this.style.paddingY) ||
-      0
+      this._ProgressBar._getTransition('alpha')._targetValue !== 0
+        ? this._ProgressBar._getTransition('y')._targetValue ||
+          this._ProgressBar.y
+        : 0) || 0
```
anthony9187, line 365: could we comment this? I'm just trying to read it and barely able to wrap my head around all the nested-ness

erautenberg: ha for sure 😅 updated with a big comment block, hopefully it helps!

erautenberg:
@ImCoolNowRight yep, that's intentional

anthony9187 (approved):

PR #234, opened 2023-06-02
fix(Button): remove unnecessary utils call and return calculated text width

erautenberg:
## Description

When dealing with fonts with custom fontStyle set, the `measureTextWidth` util method acts unexpectedly. It does not seem to actually be needed for the Button anyway.

## Checklist

- [ ] all commented code has been removed
- [ ] any new console issues have been resolved
- [ ] code linter and formatter has been run
- [ ] test coverage meets repo requirements
- [ ] PR name matches the expected semantic-commit syntax

PR #242, opened 2023-06-08
feat(TestRenderer): allow configuring stage in unit tests application

THoj13:
## Description
This change resolves a failing unit test in this PR https://github.com/rdkcentral/Lightning-UI-Components/pull/234. The test references the width of a Text texture. The calculation for that value uses `stage.precision`, so can have different results based on that value.

This change allows configuring the `stage` property on the Lightning Application returned by `TestRenderer.create`. The stage can be configured for unit tests in the following ways:
1. `TestUtils.makeCreateComponent`: The third argument in this function accepts a default options object. Set a custom `stage` on the stage property in this object. This function returns a function which can be invoked to return an instance of the component and the `testRenderer` object. The stage set by this method will be used in all tests that use the function returned by it.
2. `TestRenderer.create`: This function accepts an object as its second argument. `stage` can be defined in that options object. This will return a `testRenderer` which will refer to a Lightning Application using the defined stage.

If no `stage` is passed in the options object, the default stage will be used:
```js
const stage = {
  w: 1920,
  h: 1080,
  precision: 2 / 3,
  fontSharp: {
    precision: 2 / 3,
    fontSize: 12
  },
  useImageWorker: false
};
```
Passing in a stage to options will overwrite this default stage, not merge with it.

## Testing
1. run unit tests, everything should pass
Extra: try configuring stage in a unit test file using the instructions above

## Automation
n/a

## Checklist


packages/@lightningjs/ui-components/src/components/Marquee/Marquee.test.js
```diff
@@ -179,19 +179,21 @@ describe('Marquee', () => {
...
+          h: 300,
+          centerAlign: true
+        },
+        { spyOnMethods: ['_positionTexture'] }
+      );
+      await marquee.__positionTextureSpyPromise;
+
+      // this is the equation used to calculate x in Marqee._positionTexture when centering ContentBox
+      const centeredX = (marquee.w - marquee._textRenderedW) / 2;
+      expect(marquee._ContentBox.x).toBe(centeredX);
```
THoj13, line 196: This test failed after changing `stage.precision` to `2/3`.  Similar to the Button test failing in https://github.com/rdkcentral/Lightning-UI-Components/pull/234, I believe the calculation ultimately relies on a value internal to Lightning referencing `precision` so a change here is okay.  I updated it to instead reference the same calculation used in the component to determine the correct x value, rather than a hard coded number.

soumyaloka (approved):

packages/@lightningjs/ui-components-test-utils/src/lightning-test-renderer.js
```diff
@@ -29,8 +29,11 @@ context.setTheme({
 const stage = {
   w: 1920,
   h: 1080,
-  clearColor: 0x00000000,
-  canvas2d: false,
+  precision: 2 / 3,
```
erautenberg, line 32: I'm a little confused by the default here. I thought if the precision is 2/3, then the w/h should be 1280x720 (like our Storybook setup). Otherwise, if you actually try testing this setup in Storybook, you'll see that the stage size is full size, but the content on it is shrunk down.

THoj13: That makes sense for both to match. Updated the TestRenderer stage to match what is used in the storybook config here: 8786de0ed99bfe5c58b843dd84a57b1e3ab2deea

erautenberg (approved):

PR #243, opened 2023-06-09
docs(FocusManager): remove unused properties and add new methods

soumyaloka:
## Description

In this PR, did some some cleanup, remove unused properties from the properties table and updated existing methods, and also added a few new methods.

## References

LUI-868

## Testing

Go through the documentation and look for any spelling errors, any suggestions for better explanations and if there are any methods that I may have missed adding to the documentation

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.mdx
+++ b/packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.mdx
@@ -119,16 +119,22 @@ class FancyFocus extends FocusManager {
 
 ### Properties
 
-| name         | type            | default | description                                                        |
-| ------------ | --------------- | ------- | ------------------------------------------------------------------ |
-| direction    | 'column'\|'row' | false   | Determines the key handling for focus ('left/right' or 'up/down'). |
-| wrapSelected | boolean         | false   | Enables wrapping behavior for focus navigation.                    |
+| name          | type            | default   | description                                                    |
+| ------------- | --------------- | --------- | -------------------------------------------------------------- |
+| direction     | 'column'\|'row' | undefined | The navigation direction for focus ('left/right' or 'up/down') |
+| selectedIndex | number          | undefined | The index of the currently selected item.                      |
+| Items         | FocusItemsType  | undefined | Child element or elements of the FocusManager.                 |
+| wrapSelected  | boolean         | false     | Enables wrapping behavior for focus navigation.                |
 
 ### Methods
 
-#### appendItems(items: <DocsLink id="lng.Component">lng.Component[]</DocsLink>, idx: number): void
+#### appendItems(items: <DocsLink id="lng.Component">lng.Component[]</DocsLink>): void
 
-Appends items to child list and updates focus
+This method appends the provided rows to the end of the FocusManager's children
+
+#### appendItemsAt(items: <DocsLink id="lng.Component">lng.Component[]</DocsLink>, idx: number): void
+
+This method is similar to appendItems, but it allows appending items at a specific index in the list of child items.
 
 ##### Arguments
 
@@ -139,7 +145,7 @@ Appends items to child list and updates focus
 
 #### prependItems(items: <DocsLink id="lng.Component">lng.Component[]</DocsLink>): void
 
-Prepends items to the child list and updates the focus
+This method prepends items to the beginning of the FocusManager's children.
 
 | name  | type                                                    | required | default | description                 |
 | ----- | ------------------------------------------------------- | -------- | ------- | --------------------------- |
@@ -147,24 +153,24 @@ Prepends items to the child list and updates the focus
 
 #### removeItemAt(index: number): void
 
-Removes an item at a specific index from the child list and updates the focus
+Removes an item at a specific index from the FocusManager's children.
+
+#### TransitionXTargetValue(): string
+
+Returns the transition value for the x coordinate of the FocusManager's items.
+
+#### TransitionYTargetValue(): string
+
+Returns the transition value for the y coordinate of the FocusManager's items.
 
 #### render(): void
 
 A no-op function that is called when `selectedIndex` is set. Can be overridden by classes that extend `FocusManager` for custom render behavior.
 
-#### selected():<DocsLink id="lng.Component">lng.Component</DocsLink>
-
-Returns the currently focused item
-
 #### selectPrevious(): void
 
-Selects previous item. If this.wrapSelected=true, will select the last element in the list if focus is currently on the first item.
+Selects the previous item in the FocusManager's children. If wrapSelected is true, it selects the last item if the focus is currently on the first item.
 
 #### selectNext(): void
 
-Selects next item. If this.wrapSelected=true, will select the first element in the list if focus is currently on the last item.
-
-#### selectedIndex(): number
-
-Returns the index of the currently focused item.
+Selects the next item in the FocusManager's children. If wrapSelected is true, it selects the first item if the focus is currently on the last item.
--- a/packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.mdx
+++ b/packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.mdx
@@ -119,12 +119,12 @@ class FancyFocus extends FocusManager {
 
 ### Properties
 
-| name          | type            | default   | description                                                    |
-| ------------- | --------------- | --------- | -------------------------------------------------------------- |
-| direction     | 'column'\|'row' | undefined | The navigation direction for focus ('left/right' or 'up/down') |
-| selectedIndex | number          | undefined | The index of the currently selected item.                      |
-| Items         | FocusItemsType  | undefined | Child element or elements of the FocusManager.                 |
-| wrapSelected  | boolean         | false     | Enables wrapping behavior for focus navigation.                |
+| name          | type                                                    | default   | description                                                    |
+| ------------- | ------------------------------------------------------- | --------- | -------------------------------------------------------------- |
+| direction     | 'column'\|'row'                                         | undefined | The navigation direction for focus ('left/right' or 'up/down') |
+| selectedIndex | number                                                  | undefined | The index of the currently selected item.                      |
+| items         | <DocsLink id="lng.Component">lng.Component[]</DocsLink> | undefined | Child element or elements of the FocusManager.                 |
+| wrapSelected  | boolean                                                 | false     | Enables wrapping behavior for focus navigation.                |
 
 ### Methods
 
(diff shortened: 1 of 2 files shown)


erautenberg (approved):
Looks good, just would like to make sure you and @jazzyw coordinate since she's been updating the TS files.

ImCoolNowRight (approved):

packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.mdx
```diff
@@ -119,35 +119,52 @@ class FancyFocus extends FocusManager {
 
 ### Properties
```
jazzyw, line 120: @erautenberg In FocusManager.js, we have getters and setters for props like `selectedIndex`, `items`, `itemPosX`, `itemPosY`. They're not part of the properties array but are specific to FocusManager and can be set/changed and I believe patched to the component thus included in the TS file. That said, should they be included in this properties table?

packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.mdx
```diff
@@ -119,35 +119,52 @@ class FancyFocus extends FocusManager {
...
 
-| name          | type                                                    | readonly | description                                                                                                                  |
-| ------------- | ------------------------------------------------------- | -------- | ---------------------------------------------------------------------------------------------------------------------------- |
-| direction     | 'column'\|'row'                                         | false    | determines whether focus handling responds to left/right or up/down key                                                      |
-| selected      | <DocsLink id="lng.Component" />                         | true     | current item in focus                                                                                                        |
```
jazzyw, line 125: Similarly, `selected` only has a getter which makes the prop read-only so this was correctly documented but is it necessary in the props table? @erautenberg

packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.mdx
```diff
@@ -119,35 +119,52 @@ class FancyFocus extends FocusManager {
...
-| loading       | boolean                                                 | true     | False if an item is in focus, true otherwise                                                                                 |
-| items         | <DocsLink id="lng.Component">lng.Component[]</DocsLink> | false    | alias for `children`                                                                                                         |
-| wrapSelected  | boolean                                                 | false    | enables wrapping behavior, so `selectNext` selects the first item if the current item is the last on the list and vice versa |
+| name         | type            | default | description                                                        |
+| ------------ | --------------- | ------- | ------------------------------------------------------------------ |
+| direction    | 'column'\|'row' | false   | Determines the key handling for focus ('left/right' or 'up/down'). |
+| wrapSelected | boolean         | false   | Enables wrapping behavior for focus navigation.                    |
 
 ### Methods
```
jazzyw, line 129: `FocusManager` has a couple more public methods that looks like haven't been added yet: `appendItemsAt()`, `getTransitionXTargetValue()`, `getTransitionYTargetValue()`

packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.mdx
```diff
@@ -119,35 +119,52 @@ class FancyFocus extends FocusManager {
...
 
-#### appendItems
+#### appendItems(items: <DocsLink id="lng.Component">lng.Component[]</DocsLink>, idx: number): void
```
jazzyw: The `appendItems` method only has `items` as a parameter. The `appendItemsAt` is a separate method that includes both `items` and `idx`.

packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.mdx
```diff
@@ -119,35 +119,52 @@ class FancyFocus extends FocusManager {
...
+#### removeItemAt(index: number): void
+
+Removes an item at a specific index from the child list and updates the focus
+
+#### render(): void
 
 A no-op function that is called when `selectedIndex` is set. Can be overridden by classes that extend `FocusManager` for custom render behavior.
 
-#### selectPrevious
+#### selected():<DocsLink id="lng.Component">lng.Component</DocsLink>
```
jazzyw: A question for @erautenberg, do we include accessors in the documentation?

packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.mdx
```diff
@@ -119,35 +119,52 @@ class FancyFocus extends FocusManager {
...
+
+| name  | type                                                    | required | default | description                 |
+| ----- | ------------------------------------------------------- | -------- | ------- | --------------------------- |
+| items | <DocsLink id="lng.Component">lng.Component[]</DocsLink> | false    | []      | The list of items to append |
+
+#### removeItemAt(index: number): void
+
+Removes an item at a specific index from the child list and updates the focus
+
+#### render(): void
```
jazzyw: Out of curiosity, `render` is a private method in `FocusManager` but it's documented here. Is this expected? My thought was only public methods are listed/described in the documentation. Is `render` a special case?

@erautenberg

packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.mdx
jazzyw: The `FocusManager.d.ts` file has descriptions for these methods which can be used in this documentation

packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.mdx
erautenberg: It looks like this needs to be fixed in the TS as well to be underscored then. I don't have a problem documenting it, but it should have the appropriate name.

packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.mdx
```diff
@@ -119,35 +119,58 @@ class FancyFocus extends FocusManager {
...
 
 ##### Arguments
 
-| name  | type                                                    | required | default | description     |
-| ----- | ------------------------------------------------------- | -------- | ------- | --------------- |
-| items | <DocsLink id="lng.Component">lng.Component[]</DocsLink> | false    | []      | items to append |
+| name  | type                                                    | required | default | description                   |
+| ----- | ------------------------------------------------------- | -------- | ------- | ----------------------------- |
+| items | <DocsLink id="lng.Component">lng.Component[]</DocsLink> | false    | []      | The list of items to append   |
+| idx   | number                                                  | true     | []      | The index to insert the items |
```
jazzyw: The default for `idx` should be `undefined` since it has a type of number rather than an empty array

packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.mdx
```diff
@@ -119,35 +119,58 @@ class FancyFocus extends FocusManager {
...
-| items         | <DocsLink id="lng.Component">lng.Component[]</DocsLink> | false    | alias for `children`                                                                                                         |
-| wrapSelected  | boolean                                                 | false    | enables wrapping behavior, so `selectNext` selects the first item if the current item is the last on the list and vice versa |
+| name          | type            | default   | description                                                    |
+| ------------- | --------------- | --------- | -------------------------------------------------------------- |
+| direction     | 'column'\|'row' | undefined | The navigation direction for focus ('left/right' or 'up/down') |
+| selectedIndex | number          | undefined | The index of the currently selected item.                      |
+| Items         | FocusItemsType  | undefined | Child element or elements of the FocusManager.                 |
```
jazzyw: I believe for this property, the array of child elements is lower case `items`.

Rather than having `FocusItemsType` (without defining what that is), we can use the original `<DocsLink id="lng.Component">lng.Component[]</DocsLink>` as the type unless you'd like to define what `FocusItemsType` is in the docs.

packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.mdx
```diff
@@ -119,35 +119,58 @@ class FancyFocus extends FocusManager {
...
+
+#### removeItemAt(index: number): void
+
+Removes an item at a specific index from the FocusManager's children.
+
+#### TransitionXTargetValue(): string
+
+Returns the transition value for the x coordinate of the FocusManager's items.
+
+#### TransitionYTargetValue(): string
```
jazzyw: Looks like these two transition methods are missing 'get' in the beginning: `getTransitionXTargetValue`, `getTransitionYTargetValue`

packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.mdx
```diff
@@ -119,35 +119,58 @@ class FancyFocus extends FocusManager {
...
+
+Returns the transition value for the y coordinate of the FocusManager's items.
 
-#### render
+#### render(): void
```
jazzyw: Emily said it's ok to include this in the docs but could you add the `_` in the beginning of the function name to match how it's named in the component file?

Likewise, if you could add the `_` to `render` in the .d.ts file too that would be awesome. Thank you!

soumyaloka: Yep, I can do that. Thanks!

packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.mdx
soumyaloka: Ok, then I would prefer to use now `<DocsLink id="lng.Component">lng.Component[]</DocsLink> `

packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.mdx
soumyaloka: What do guys think about this? @erautenberg @jazzyw

packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.mdx
```diff
@@ -119,35 +119,58 @@ class FancyFocus extends FocusManager {
...
+#### getTransitionXTargetValue(): string
+
+Returns the transition value for the x coordinate of the FocusManager's items.
+
+#### getTransitionYTargetValue(): string
+
+Returns the transition value for the y coordinate of the FocusManager's items.
 
-#### render
+#### \_render(): void
```
jazzyw, line 166: Thanks for the update! Could the slash in the beginning get removed?

soumyaloka: Linter automatically does that when I save the file, I think I should disable it and push the change

packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.mdx
jazzyw: We do document accessors in the TypeScript documentation. In my opinion it could be helpful to add accessors in the component docs. 

This addition could also be a separate ticket as all the current docs do not document accessors and will have to be updated.

packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.mdx
jazzyw: Ah I see seems like the slash isn't showing up in the actual documentation though so if the linter continues to add it I think it's fine

jazzyw (approved):

PR #245, opened 2023-06-14
fix(Tile): add metadatalocation as style property

soumyaloka:
## Description

In this PR, Added metadataLocation as a property to the style object. Updated the getter method to return the value of this.style.metadataLocation if it exists, or fallback to the previous metadataLocation value. Updated the setter method metadataLocation(newMetadataLocation) to set the value of this.style.metadataLocation to the provided newMetadataLocation, only if it is non-empty.

## References

LUI-852

## Testing

Navigate to Tile and make sure the metadatalocation control works as expected


packages/@lightningjs/ui-components/src/components/Tile/Tile.d.ts
```diff
@@ -24,6 +24,7 @@ import Surface, { SurfaceStyle } from '../Surface';
 export type TileStyle = SurfaceStyle & {
   animationEntrance: Record<string, unknown>;
   animationExit: Record<string, unknown>;
+  metadataLocation: 'standard' | 'inset';
   paddingX: number | string;
   paddingY: number | string;
   paddingYProgress: number | string;
```
erautenberg: I just realized I never added that new paddingYBetweenContent prop here 🤦‍♀️ would you mind adding it while you're in here please?

packages/@lightningjs/ui-components/src/components/Tile/Tile.js
```diff
@@ -502,6 +502,16 @@ export default class Tile extends Surface {
     };
   }
 
+  get metadataLocation() {
+    return this.style.metadataLocation ?? this.metadataLocation;
+  }
+
+  set metadataLocation(newMetadataLocation) {
```
erautenberg: I think this and the above should actually be `_setMetadataLocation` and `_getMetadataLocation` because those are the methods that the `withUpdates` mixin creates.

erautenberg: Checkout out ProgressBar for an example of overriding the props from the properties array:
```js
  _setProgress(progress) {
    this._progressChanged = progress !== this._progress;
    return progress;
  }
```

packages/@lightningjs/ui-components/src/components/Tile/Tile.js
```diff
@@ -502,6 +502,16 @@ export default class Tile extends Surface {
...
   }
 
+  _getMetadataLocation() {
+    return this.style.metadataLocation ?? this.metadataLocation;
+  }
+
+  _setMetadataLocation(newMetadataLocation) {
+    if (newMetadataLocation) {
+      this.style.metadataLocation = newMetadataLocation;
+    }
```
erautenberg, line 512: I think you still need to `return newMetadataLocation` at the end.

packages/@lightningjs/ui-components/src/components/Tile/Tile.js
```diff
@@ -502,6 +502,16 @@ export default class Tile extends Surface {
     };
   }
 
+  _getMetadataLocation() {
+    return this.style.metadataLocation ?? this.metadataLocation;
```
erautenberg: I think the second half should be this._metadataLocation

ImCoolNowRight:
Right now, if I set `persistentMetadata` to true and `mode` to unfocused, the metadata always remains under the Tile, regardless of if I set the location to standard or inset. Is this correct, or should it remain inside the Tile when set to inset?

soumyaloka:
> Right now, if I set `persistentMetadata` to true and `mode` to unfocused, the metadata always remains under the Tile, regardless of if I set the location to standard or inset. Is this correct, or should it remain inside the Tile when set to inset?

Yes that is a bug but it was not introduced with this changes. Emily found that issue on Tile component last week and she had a fix for it and here is the PR : https://github.com/rdkcentral/Lightning-UI-Components/pull/246

Here is the slack thread:https://cim.slack.com/archives/C044CH7REP2/p1686841324001869

arwehrman (approved):

ImCoolNowRight (approved):

PR #246, opened 2023-06-15
fix(Tile): re-enabled persistentMetadata and clean up transitions

erautenberg:
## Description

The inset persistent metadata was broken and after putting it back, the animation on/off the tile was happening on focus/unfocus even if the metadata should stay visible.

## Testing

1. Go to Tile
2. In standard metadata position, test that it appears/disappears on focus/unfocus appropriately.
3. Repeat inset.
2. In standard metadata position, turn on persistentMetadata and test that it is visible in all modes.
3. Repeat with inset.
4. Repeat all of the above in the circle mode. The inset metadata should never be visible (regardless of persistency).
5. Reset back to the original args, turn on persistent metadata and confirm that in both the standard and inset positions, if you remove all of the text in title and description, they are not visible on the tile (this was an issue with the short circuit return in updateMetadata not actually alpha-ing anything out).

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Tile/Tile.test.js
+++ b/packages/@lightningjs/ui-components/src/components/Tile/Tile.test.js
@@ -433,19 +433,16 @@ describe('Tile', () => {
     expect(tile._ProgressBar).toBeUndefined();
   });
 
-  it('should update metadata and remove if no longer needed if metadataLocation is inset', async () => {
+  it('should update metadata and remove if no longer needed', async () => {
     expect(tile.metadata).toBeUndefined();
+
     tile.metadata = { title: 'test' };
-    tile.metadataLocation = 'inset';
     await tile.__updateMetadataSpyPromise;
     expect(tile.metadata).not.toBeUndefined();
+
     tile.metadata = undefined;
     await tile.__updateMetadataSpyPromise;
     expect(tile._Metadata).toBeUndefined();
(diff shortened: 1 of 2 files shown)


arwehrman (approved):

packages/@lightningjs/ui-components/src/components/Tile/Tile.test.js
```diff
@@ -433,19 +433,16 @@ describe('Tile', () => {
...
     expect(tile.metadata).toBeUndefined();
+
     tile.metadata = { title: 'test' };
-    tile.metadataLocation = 'inset';
     await tile.__updateMetadataSpyPromise;
     expect(tile.metadata).not.toBeUndefined();
+
     tile.metadata = undefined;
     await tile.__updateMetadataSpyPromise;
     expect(tile._Metadata).toBeUndefined();
```
THoj13, line 446: I think adding the opposite of this assertion before setting `tile.metadata = undefined;` would be good here so that we're certain `tile._Metadata` was defined at the start of the test and was only removed after `tile.metadata` was set to `undefined`.
```js
expect(tile._Metadata).not.toBeUndefined()
```

THoj13 (commented):
Looks good to me. I left one comment for a "nice-to-have" addition to a unit test if it seems like a decent change to you, but that shouldn't block this.

THoj13 (approved):

PR #249, opened 2023-06-20
docs: add documentation for @lightningjs/ui-test-utils

THoj13:
## Description
Add documentation about what is available from the `@lightningjs/ui-test-utils` package.
- TestUtils
- TestRenderer

Utilities excluded from this round of documentation:
- `pathToDateURI`: This function doesn't currently work. The path passed in won't lead to anything unless that same asset is in both @lightningjs/tests-utils as well as the package from where the test is called from. 
- `TestRenderer.getContext`: There isn't any usage of this utility in the open or inner source repos.
- `TestRenderer.destroy`: This is only used in an afterEach in `Keyboard.test.js`, but I'm not sure there's a practical use for it.
- `TestRenderer.focusPath`: There's no usage of this in LUI.

## References
LUI-871

## Testing
1. Run Storybook
2. Read through the documentation in the "Unit Testing" Category
3. Provide feedback:
    - Do the page titles/order make sense? 
    - Is the documentation understandable? 
    - Are the examples helpful?
    - Anything else

## Automation
n/a

## Checklist


ImCoolNowRight (approved):
Absolutely fantastic! I honestly don't think anything needs to be changed. Everything is thorough and makes sense as it is!

erautenberg (approved):
This is amazing!!! Thank you so much for being so thorough. Only nit-pick is can we move the Unit Testing section up to be under the "Docs" section? Otherwise I'm just worried it'll be too easily missed.

THoj13:
> This is amazing!!! Thank you so much for being so thorough. Only nit-pick is can we move the Unit Testing section up to be under the "Docs" section? Otherwise I'm just worried it'll be too easily missed.

That's a good call, I'll update it to be in that section.

PR #254, opened 2023-06-21
fix(ProgressBar): update Base theme default brand tone barColor to neutral

erautenberg:
## Description

Per the weekly component reviews with design, we are updating the default barColor for the brand tone to use a neetral color to better match the industry standard (brand color is only used on the foreground progress bar).

## Testing

In the ProgressBar story, observe that when in the brand tone, the brand color is only applied to the progress color, and the bar color uses the neutral color.

## Checklist


soumyaloka (approved):
Looks good!

ImCoolNowRight (approved):

PR #262, opened 2023-06-28
feat: add TypeScript definitions for @lightningjs/ui-components-test-utils exports

THoj13:
## Description

- adds TypeScript definition files for all exports from the `@lightningjs/ui-components-test-utils` package

## References
LUI-872

## Automation
n/a

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components-test-utils/src/lightning-test-utils.d.ts
+++ b/packages/@lightningjs/ui-components-test-utils/src/lightning-test-utils.d.ts
@@ -44,7 +44,7 @@ type createComponent = (
 ) => [lng.Element, testRenderer];
 
 export function makeCreateComponent(
-  type: Base.TemplateSpec, // TODO: this throws TS error when used
+  type: lng.Component.Constructor,
   defaultConfig?: makeCreateComponentConfig,
   defaultOptions?: makeCreateComponentDefaultOptions
 ): createComponent;
--- a/packages/@lightningjs/ui-components-test-utils/src/lightning-test-utils.d.ts
+++ b/packages/@lightningjs/ui-components-test-utils/src/lightning-test-utils.d.ts
@@ -17,10 +17,10 @@
  */
 import lng from '@lightningjs/core';
 import { jest } from '@jest/globals';
-import type {
+import {
   default as TestRenderer,
   testRenderer
-} from './lightning-test-renderer.d.ts';
+} from './lightning-test-renderer';
 
 /**
  * nextTick
@@ -91,7 +91,7 @@ export function completeAnimation(
  * mockContext
  */
 // TODO: no TS def for Context available
-type context = Record<string, unknown>;
+export type context = Record<string, unknown>;
 
 export function mockContext(
   context: context,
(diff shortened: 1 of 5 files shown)


packages/@lightningjs/ui-components-test-utils/src/lightning-test-utils.d.ts
```diff
@@ -0,0 +1,70 @@
...
+import type {
+  default as TestRenderer,
+  testRenderer
+} from './lightning-test-renderer.d.ts';
+
+export function nextTick(wait?: number): Promise<void>;
+
+export function fastForward(elements: lng.Element[]): void;
+
+type makeCreateComponentConfig = Record<string, any>;
```
anthony9187: can this be `Record<string, unknown>` or does it need to be `any`?

anthony9187 (commented):
we should make a habit of adding an eslint disable line wherever we're using `any`. I know it's a bit annoying, but we should be explicit about why we're not providing a type in these instances

packages/@lightningjs/ui-components-test-utils/src/lightning-test-utils.d.ts
THoj13: Ya, switching it to `unknown` should be fine.

THoj13:
> we should make a habit of adding an eslint disable line wherever we're using any. I know it's a bit annoying, but we should be explicit about why we're not providing a type in these instances 

That makes sense. I'll go through and check those. There's likely other cases where `any` can be switched to `unknown` like you caught in your other comment on `makeCreateComponent`.

THoj13:
~There's some TS errors I'm getting when testing out makeCreatComponent and TestRenderer.create. Marking this as a draft until those are resolved.~

These are resolved now.

erautenberg (approved):

packages/@lightningjs/ui-components-test-utils/src/lightning-test-utils.d.ts
```diff
@@ -0,0 +1,117 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import lng from '@lightningjs/core';
+import { jest } from '@jest/globals';
+import type {
```
anthony9187: we can probably omit the `type` keyword here

packages/@lightningjs/ui-components-test-utils/src/lightning-test-utils.d.ts
```diff
@@ -0,0 +1,117 @@
...
+export function completeAnimation(
+  element: lng.Element,
+  transitionProperties?: string | string[]
+): Promise<void>;
+
+/**
+ * mockContext
+ */
+// TODO: no TS def for Context available
+type context = Record<string, unknown>;
```
anthony9187: we may need to export this since it's referenced in an exported function

anthony9187 (approved):
👍

PR #263, opened 2023-06-29
fix: update alpha handling in InlineContent Truncation Logic

soumyaloka:
## Description

In this PR, this change ensures the alpha value is changed only when truncation is required, so won’t affect Marquee setting its own alpha on InlineContent.

## References

Part of LUI-889

## Testing

Verify inlineContent stories haven't effected and marquee sub story centeredText should work as expected.


ImCoolNowRight (approved):
Everything looks good!

THoj13 (approved):

PR #264, opened 2023-06-30
fix(CI): add git username

ImCoolNowRight:
## Description

Adds git username to avoid error when back merging during release job.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -47,7 +47,7 @@ jobs:
       - name: Setup git config
         run: |
           # setup the username and email
-          git config user.name "GitHub Actions Bot"
+          git config user.name "svc-lightning-ui-components"
           git config user.email "<>"
       - name: Save changes
         run: |


anthony9187:
I'm curious, what's the error we're resolving here? the specific git username shouldn't really matter since permissions are based on the PAT. is it mad about the spaces?

erautenberg (approved):

joshhowenstine:
@anthony9187 I found that when testing the merge fails if the username does not match the username that the PAT was created with. We may be able to remove the config email however @ImCoolNowRight.

ImCoolNowRight:
> I'm curious, what's the error we're resolving here? the specific git username shouldn't really matter since permissions are based on the PAT. is it mad about the spaces?

@anthony9187 This is the error we were encountering during the build: https://github.com/rdkcentral/Lightning-UI-Components/actions/runs/5349179450/jobs/9700088743

anthony9187 (approved):

PR #265, opened 2023-07-03
fix(ListItem): correct title style applied during truncation

ImCoolNowRight:
## Description

Corrects the style that is being applied in `_updateTruncation` for the Title.

## Testing

Add custom styles for the Title TextBox in the ListItem story. You should see that the Title updates correctly now.

## Checklist


soumyaloka (approved):
Good catch, Thanks for updating this.

packages/@lightningjs/ui-components/src/components/ListItem/ListItem.js
```diff
@@ -111,7 +111,17 @@ export default class ListItem extends Button {
   }
 
   _updateTruncation() {
-    super._updateTruncation();
+    if (this._Title) {
```
joshhowenstine, line 114: Does this work if moved to _updateTitle? Just wondering if it makes more sense to live in that method if possible.

ImCoolNowRight: Tested it out and it doesn't work if I move it out of here.

joshhowenstine (approved):

PR #274, opened 2023-07-07
feat: alias support for class properties and style properties

erautenberg:
## Description

These updates add aliasing support to withUpdates and withThemeStyles so that we can update property names and throw deprecation warnings without creating breaking changes. This also adds support for `height` and `width` to be accepted as `h` and `w` in the styles file.

Once this is vetted and merged, I will be going through each component to scrub for abbreviated properties like "iconW" and replacing them with the expanded string version ("iconWidth") using this new alias feature, both in the component properties and component style properties.

## Testing

1. Add "value" as a deprecated property in ProgressBar
    - In Progress.d.ts's class declaration (note that these "strings" need to be changed to "number" when we scrub the components):
      ```ts
      progress: string;
      /** @deprecated */
      value: string;
      ```
    - In Progress.js:
      ```js
      static get aliasProperties() {
        return [{ prev: 'value', curr: 'progress' }];
      }
      ```
2. Add "foregroundColor" as a deprecated style property in ProgressBar
    - In Progress.d.ts's ProgressBarStyle declaration:
      ```ts
      progressColor: Color;
      /** @deprecated */
      foregroundColor: Color;
      ```
    - In Progress.js
      ```js
      static get aliasStyles() {
        return [{ prev: 'foregroundColor', curr: 'progressColor' }];
      }
      ```
3. Add instructions to ProgressBar story to test out the aliasing:
    - In ProgressBar.stories.js, under the template method, add something like this:
      ```js
      _init() {
        setTimeout(() => {
          this.tag('ProgressBar').value = 50;
          this.tag('ProgressBar').style = { foregroundColor: 0xff5b16e5 };
        }, 3000);
      }
      ```
4. Navigate to the ProgressBar story in the browser and open the console. Observe that after 3 seconds, the progress bar value (progress) is changed along with the progress bar color (progressColor) and there are warnings in the console for both the "value" and "foregroundColor" properties, like below:
    Screen Shot 2023-07-07 at 4 03 34 PM

5. If able to test in a TS file, you will notice that setting the deprecated properties will display a strikethrough and warning on hover:
    Screen Shot 2023-07-07 at 2 46 16 PM

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/mixins/withThemeStyles/index.js
+++ b/packages/@lightningjs/ui-components/src/mixins/withThemeStyles/index.js
@@ -148,13 +148,13 @@ export default function withThemeStyles(Base, mixinStyle) {
     }
 
     get _modeStyle() {
-      return this.replaceAliasValues(
+      return this._replaceAliasValues(
         this._componentStyleSource?.mode?.[this.mode] || {}
       );
     }
 
     get _toneStyle() {
-      return this.replaceAliasValues(
+      return this._replaceAliasValues(
         this._componentStyleSource?.tone?.[this.tone] || {}
       );
     }
@@ -164,7 +164,7 @@ export default function withThemeStyles(Base, mixinStyle) {
      * @return {object}
      */
     get _themeLevelStyle() {
-      return this.replaceAliasValues(this._componentConfig?.style || {});
+      return this._replaceAliasValues(this._componentConfig?.style || {});
     }
 
     /**
@@ -172,7 +172,7 @@ export default function withThemeStyles(Base, mixinStyle) {
      * @return {object}
      */
     get _componentLevelStyle() {
-      return this.replaceAliasValues(this._componentLevelStyleSource || {});
+      return this._replaceAliasValues(this._componentLevelStyleSource || {});
     }
 
     /**
@@ -460,7 +460,7 @@ export default function withThemeStyles(Base, mixinStyle) {
       this.queueThemeUpdate();
     }
 
-    replaceAliasValues(value) {
+    _replaceAliasValues(value) {
       const styleObj = clone(value, {});
       const aliasProps = [
         { prev: 'height', curr: 'h', skipWarn: true },
--- a/packages/@lightningjs/ui-components/src/mixins/withThemeStyles/index.js
+++ b/packages/@lightningjs/ui-components/src/mixins/withThemeStyles/index.js
@@ -476,7 +476,7 @@ export default function withThemeStyles(Base, mixinStyle) {
         ) {
           !alias.skipWarn &&
             console.warn(
-              `The style property "${alias.prev}" is deprecated. Please use "${alias.curr}" instead.`
+              `The style property "${alias.prev}" is deprecated and may be removed in a future release. Please use "${alias.curr}" instead.`
             );
           Object.defineProperty(
             styleObj,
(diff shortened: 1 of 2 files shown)


packages/@lightningjs/ui-components/src/mixins/withThemeStyles/index.js
```diff
@@ -456,6 +460,35 @@ export default function withThemeStyles(Base, mixinStyle) {
       this.queueThemeUpdate();
     }
 
+    replaceAliasValues(value) {
```
erautenberg: Didn't really think about it at the time, do you all have a preference for if this is a "private" (underscored) method or not?

joshhowenstine: Yeah, prob could be underscored.

anthony9187: +1 for private

packages/@lightningjs/ui-components/src/mixins/withThemeStyles/index.js
```diff
@@ -456,6 +460,35 @@ export default function withThemeStyles(Base, mixinStyle) {
...
+      aliasProps.forEach(alias => {
+        if (
+          alias &&
+          typeof alias.prev === 'string' &&
+          typeof alias.curr === 'string' &&
+          styleObj[alias.prev]
+        ) {
+          !alias.skipWarn &&
+            console.warn(
+              `The style property "${alias.prev}" is deprecated. Please use "${alias.curr}" instead.`
```
anthony9187: should we add something like "and will be removed in a future version"? idk if that's implied or if being more explicit is better

packages/@lightningjs/ui-components/src/mixins/withUpdates/index.js
```diff
@@ -49,6 +49,20 @@ function getPropertyDescriptor(name, key) {
   };
 }
 
+function getAliasPropertyDescriptor(prev, curr) {
+  const deprecationWarning = `The property "${prev}" is deprecated. Please use "${curr}" instead.`;
```
anthony9187: same here as above - should we add something like "and will be removed in a future version"? idk if that's implied or if being more explicit is better

packages/@lightningjs/ui-components/src/mixins/withThemeStyles/index.js
erautenberg: Good idea. I'd rather be as clear as possible so people can't come back and say "i DiDn'T kNoW dEpRecATing mEAnT bAAd" 🙃

packages/@lightningjs/ui-components/src/mixins/withThemeStyles/index.js
erautenberg: updated!

packages/@lightningjs/ui-components/src/mixins/withThemeStyles/index.js
erautenberg: Updated to add that "____ is deprecated and may be removed in a future release." Let me know if you think that's OK or not aggressive enough.

packages/@lightningjs/ui-components/src/mixins/withUpdates/index.js
erautenberg: updated to match the above

joshhowenstine (approved):
Looks good!

anthony9187 (approved):

PR #286, opened 2023-07-18
feat(Badge): add stroke support and clean up docs file

soumyaloka:
## Description

1. In this PR, I have added stroke functionality to the Badge component, did some cleanup in docs file and fixed the Style properties table by removing extra columns.
2. Added alpha functionality to the Badge component when there is no title or no icon present.

## References

LUI-912

## Testing

1. Navigate to Badge component.
2. Verify that the stroke support has been added.
3. Test the layout and positioning of the title and icon are working as expected
4. Try removing the title in the text sub story and make sure it should be completely alpha out(before it used to have small surface).


ImCoolNowRight:
It doesn't look like the text is properly centered (or there's too much padding on the right side).

Screen Shot 2023-07-18 at 4 47 40 PM

soumyaloka:
> It doesn't look like the text is properly centered (or there's too much padding on the right side).
> 
> Screen Shot 2023-07-18 at 4 47 40 PM

@erautenberg @ImCoolNowRight 
I noticed that this issue is caused by the textAlign property which is not working as expected. we have two possible options:
Adjust the text's x position to center it based on the respective conditions.
Ensure that the textAlign property is properly implemented and functioning correctly(so manually need not to adjust x values).
which one would be ideal?

arwehrman (approved):

packages/@lightningjs/ui-components/src/components/Badge/Badge.js
```diff
@@ -68,29 +70,53 @@ export default class Badge extends Base {
     this._updateText();
     this._updateIcon();
     this._updateLayout();
+    this._updateVisibility();
+  }
+
+  _updateVisibility() {
+    if (!this.title && !this.icon) {
+      this.alpha = 0.001;
```
erautenberg: I think this could even be alpha 0. If neither exist, we should be OK to totally de-render it

erautenberg (commented):
Is there an advantage to having the background and the stroke separate? Since you're using the getRoundRect texture already, you should be able to specify the fill color as well:
https://github.com/rdkcentral/Lightning/blob/8378d0e69752f476abedecc85568a27dbb63cbb0/src/tools/Tools.mjs#L30

erautenberg:
> > It doesn't look like the text is properly centered (or there's too much padding on the right side).
> > Screen Shot 2023-07-18 at 4 47 40 PM
> 
> @erautenberg @ImCoolNowRight I noticed that this issue is caused by the textAlign property which is not working as expected. we have two possible options: Adjust the text's x position to center it based on the respective conditions. Ensure that the textAlign property is properly implemented and functioning correctly(so manually need not to adjust x values). which one would be ideal?

It doesn't seem that misaligned currently on in [Storybook on the release branch](https://rdkcentral.github.io/Lightning-UI-Components/?path=/story/metadata-badge--text) 🤔 Can we hunt down what caused the shift? Is it something with the new shader? Maybe the Text is in the right place, but the background isn't? Adding a random offset on top of the padding seems hacky to me.

soumyaloka:
> > > It doesn't look like the text is properly centered (or there's too much padding on the right side).
> > > Screen Shot 2023-07-18 at 4 47 40 PM
> > 
> > 
> > @erautenberg @ImCoolNowRight I noticed that this issue is caused by the textAlign property which is not working as expected. we have two possible options: Adjust the text's x position to center it based on the respective conditions. Ensure that the textAlign property is properly implemented and functioning correctly(so manually need not to adjust x values). which one would be ideal?
> 
> It doesn't seem that misaligned currently on in [Storybook on the release branch](https://rdkcentral.github.io/Lightning-UI-Components/?path=/story/metadata-badge--text) 🤔 Can we hunt down what caused the shift? Is it something with the new shader? Maybe the Text is in the right place, but the background isn't? Adding a random offset on top of the padding seems hacky to me.

I agree, adding an extra offset to fix the alignment issue does seem like a hacky solution. need to understand what caused the shift. It could be related to the new shader or a positioning problem with the background

erautenberg (approved):

PR #289, opened 2023-07-19
docs(Storybook): simplify Storybook sections to just Utilities and Components

erautenberg:
## Description

The different sections in Storybook have proven to be more confusing than helpful, so condensing all stories into Docs, Utilities  (mixins) and Components. Leaving support for potential future "Patterns" and "Templates."

## Testing

1. Confirm Storybook looks as expected with these new sections
2. Confirm the embedded canvases still work as expected (Docs tab of a component)
3. Confirm links to other components work in the docs

## Checklist


packages/@lightningjs/ui-components/src/mixins/withLayout/withLayout.mdx
```diff
@@ -109,7 +109,7 @@ If the mixin is unable to calculate the height OR width from the passed paramete
 If the mixin is unable to calculate height AND width from the passed parameters. No autosize will occur.
 
 <Canvas>
-  <Story id="utilities-withlayout--with-layout" />
+  <Story id="utilities--withlayout--with-layout" />
```
ImCoolNowRight: Not going to comment on them all, but this extra hyphen is causing all of the utilities stories to not load in the docs.

ImCoolNowRight (changes requested):

packages/@lightningjs/ui-components/src/components/Control/Control.mdx
```diff
@@ -80,7 +80,7 @@ Optional flag for the `Control` element to collapse (hide the title) when `Contr
 
 ### Style Properties
 
-`Control` contains the same style properties as [Button](?path=/docs/controls-button--button) in additional to a couple other style properties listed below:
+`Control` contains the same style properties as [Button](?path=/docs/components-button--button) in additional to a couple other style properties listed below:
```
soumyaloka, line 83: Path for the button works fine in the style properties, but it is not working in properties section, also noticed they are few other spots that needs to be updated like in TitleRow, Control and few other component docs.
I can submit another quick PR to update those.

arwehrman:
A heads up, a lot of this changes in v7.

- We can't use the variables in Titles or in story sorting. See my comments on my v7 pr about that 
- We no longer use the <Canvas> in the docs

I can make these updates in v7 but just wanted to be mindful if we do any future changes like this before v7 is merged in.

packages/@lightningjs/ui-components/src/mixins/withLayout/withLayout.mdx
erautenberg: Fixed, thanks!

erautenberg:
@arwehrman I'm not sure when we're merging that PR, so I'd like to get these in regardless to make things cleaner for AirUI. Quick find and replaces on your shouldn't be too bad anyway and I can help if you need.

> * We can't use the variables in Titles or in story sorting. See my comments on my v7 pr about that

Why can't we used variables anymore? I read your post and looked at the migration guide and understand that the sort function needs to be self-contained so we can't use variables _there,_ but the stories themselves are still just JS and should still be able to import the categories without issue.

> * We no longer use the  in the docs

What do you mean by the second bullet, looks like a word is missing?

arwehrman:
`title`  in v7 is optional and if left out it will use the files location. Also `title` can only be a string  see: https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#string-literal-titles
In v7, we could move everything into sub-folders, so Utilities, Docs, Components, and other future categories. Then SB would use that sub-folder as the category for the title automatically.

Second part about the Docs, not putting <Canvas> in the Docs.  We can, but I thought we wanted to remove them. Not super important for this PR but just something to keep in mind.

packages/@lightningjs/ui-components/src/components/Control/Control.mdx
soumyaloka: I have created a ticket for this, may be after merging this PR and Mandy's PR. That should be ready to be picked by anyone. https://ccp.sys.comcast.net/browse/LUI-939

erautenberg:
> `title` in v7 is optional and if left out it will use the files location. Also `title` can only be a string see: https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#string-literal-titles In v7, we could move everything into sub-folders, so Utilities, Docs, Components, and other future categories. Then SB would use that sub-folder as the category for the title automatically.
> 
> Second part about the Docs, not putting in the Docs. We can, but I thought we wanted to remove them. Not super important for this PR but just something to keep in mind.

Wow this is so wild. Why is Storybook making that so strict, it seems odd if it all boils down to a string in the end 🙄 Is it OK to keep this for now and in your upgrade PR we can just do a find and replace on everything or does that create too much rework?

I'm not opposed to using the folders either, the mixins are in their old folder outside of components already, so maybe for the component-specific stories, removing the path altogether is the simple solution

ImCoolNowRight (approved):

packages/@lightningjs/ui-components/src/components/Tile/Tile.mdx
```diff
@@ -90,16 +90,16 @@ class Basic extends lng.Component {
...
+| metadata           | object  | false    | undefined  | Object containing all properties supported in the [MetadataTile component](?path=/docs/components-metadatatile--metadata-tile)<br /> Can use a different Metadata component by passing in a 'type' and then that component's properties |
+| metadataLocation   | string  | false    | 'standard' | Controls where there metadata is displayed in relation to the Tile. Available values are 'standard' and 'inset'                                                                                                                         |
+| persistentMetadata | boolean | false    | false      | Metadata will be shown at all times if set to true, otherwise it will only show when the Tile has focus                                                                                                                                 |
+| progressBar        | object  | false    | undefined  | Object containing all properties supported in the [ProgressBar component](??path=/docs/utilities-progressbar--progress-bar)                                                                                                             |
```
soumyaloka: We have to change path for progressBar and Checkbox, they are pointed to utilities here.

packages/@lightningjs/ui-components/src/components/MetadataCardContent/MetadataCardContent.mdx
```diff
@@ -55,25 +55,25 @@ class Basic extends lng.Component {
...
+| name        | type                                                 | required | default   | description                                                                              |
+| ----------- | ---------------------------------------------------- | -------- | --------- | ---------------------------------------------------------------------------------------- |
+| title       | string                                               | false    | undefined | title text                                                                               |
+| description | string                                               | false    | undefined | description text directly below title                                                    |
+| details     | string                                               | false    | undefined | details text at bottom left of component                                                 |
 | provider    | [Provider](?path=/docs/utilities-provider--provider) | false    | undefined | an object of [Provider](?path=/docs/utilities-provider--provider) properties to patch in |
```
soumyaloka: Same here need to replace with components.

soumyaloka: In CardRadio, we need change the path for Radio from utilities to components

soumyaloka (approved):

PR #292, opened 2023-07-25
fix(InlineContent): default InlineContent alignment should be flex-start

erautenberg:
## Description

Updating the default flexbox alignment of the InlineContent component to be "flex-start" instead of "center" in order to better match how Lightning Text's default `textAlign` property works (default to left). Then in the TextBox component, I added a function from `utils` called `convertTextAlignToFlexJustify` that maps the left, center, and right text alignment to flex-start, center, and flex-end. I also removed the hard-coded "flex-start" assignments in Marquee and MetadataBase since they should no longer be needed and now can be overwritten as needed by the user via styles.

## Testing

Check TextBox, InlineContent, MetadataBase, MetadataCardContent, and Marquee to confirm everything is still working as expected.

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/utils/index.js
+++ b/packages/@lightningjs/ui-components/src/utils/index.js
@@ -802,6 +802,30 @@ export function watchForUpdates({
   return element;
 }
 
+/**
+ * Given a Lightning text alignment option (left, right, center),
+ * return the Lightning flexbox equivalent.
+ *
+ * @param {string} align
+ * @returns {string}
+ */
+export function convertTextAlignToFlexJustify(align) {
+  switch (align) {
+    case 'left':
+      return 'flex-start';
+    case 'center':
+      return 'center';
+    case 'right':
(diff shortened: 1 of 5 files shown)


THoj13 (approved):

packages/@lightningjs/ui-components/src/utils/index.js
```diff
@@ -802,6 +802,25 @@ export function watchForUpdates({
...
+export function convertTextAlignToFlexJustify(align) {
+  switch (align) {
+    case 'left':
+      return 'flex-start';
+    case 'center':
+      return 'center';
+    case 'right':
+      return 'flex-end';
+  }
+  return 'flex-start'; // if there is no alignment passed in, the Lightning default is "left"
```
anthony9187: is it possible use the switch statement `default` here instead of the return outside the switch?

soumyaloka (approved):
Everything works as expected!

PR #293, opened 2023-07-25
fix(ListItem): update type declarations

anthony9187:
## Description

This PR updates the ListItem type declarations to follow the pattern of subclassable components from @lightningjs/core.

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/ListItem/ListItem.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/ListItem/ListItem.d.ts
@@ -19,7 +19,7 @@
 import lng from '@lightningjs/core';
 import Button, { ButtonStyle } from '../Button';
 import { TextBoxStyle } from '../TextBox';
-import type { StylePartial } from '../../types/lui';
+import { StylePartial } from '../../types/lui';
 
 type LogoStyleObject = {
   radius: lng.Tools.CornerRadius;
@@ -27,7 +27,7 @@ type LogoStyleObject = {
   w: number;
 };
 
-export type ListItemStyle = ButtonStyle & {
+type ListItemStyle = ButtonStyle & {
   alpha: number;
   descriptionTextStyle: TextBoxStyle;
   h: number;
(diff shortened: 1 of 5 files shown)


packages/@lightningjs/ui-components/src/components/ListItem/ListItemSlider.d.ts
```diff
@@ -18,20 +18,49 @@
...
+  export interface TemplateSpec extends ListItem.TemplateSpec {
+    /**
+     * object containing all properties supported in the [Slider component](?path=/docs/elements-slider--basic)
+     */
+    slider?: Record<string, unknown>;
+
+    /**
+     * current value of slider
+     */
+    value?: string;
```
ImCoolNowRight: Shouldn't this be a number? (And changed below?)

packages/@lightningjs/ui-components/src/components/ListItem/ListItem.d.ts
```diff
@@ -19,15 +19,15 @@
...
   radius: lng.Tools.CornerRadius;
   h: number;
   w: number;
 };
 
-export type ListItemStyle = ButtonStyle & {
+type ListItemStyle = ButtonStyle & {
   alpha: number;
   descriptionTextStyle: TextBoxStyle;
   h: number;
```
soumyaloka: I think we can remove the 'h' style

packages/@lightningjs/ui-components/src/components/ListItem/ListItemSlider.d.ts
anthony9187: it absolutely should be

packages/@lightningjs/ui-components/src/components/ListItem/ListItemPicker.d.ts
```diff
@@ -30,10 +30,41 @@ export type ListItemPickerStyle = ListItemStyle & {
...
+  /**
+   * index of current selected option
+   */
   selectedIndex?: number;
-  get selectedOption(): Record<string, unknown>; // TODO is this correct?
+
+  /**
+   * returns the item from the `options` array at the current `selectedIndex`
+   */
+  get selectedOption(): string; // TODO is this correct?
```
soumyaloka: This looks correct to me, but I'm not sure if we should include this getter in the class export

ImCoolNowRight (approved):

packages/@lightningjs/ui-components/src/components/ListItem/ListItemPicker.d.ts
anthony9187: if it's a public method(not underscored) it has to be included. the thing that's getting flagged here I think, is whether or not that return type is correct

soumyaloka: return type 'string' for the selectedOption is correct, since options is an array of strings and selectedOption will retrieve a single selected item from the options.
 Does that make sense?

soumyaloka (approved):

packages/@lightningjs/ui-components/src/components/ListItem/ListItemPicker.d.ts
ImCoolNowRight: Looks like it's actually an object that represents the whole item being selected. Not sure if that's what we want as the result of that getter or if we wanted something more narrow (like its title), but for now it's an object.

packages/@lightningjs/ui-components/src/components/ListItem/NestedSlider.d.ts
```diff
@@ -24,7 +25,10 @@ export type NestedSliderStyle = SliderStyle & {
   innerCircleSize: number;
```
soumyaloka: we are not using innerCircleSize anymore, we can delete this.

packages/@lightningjs/ui-components/src/components/ListItem/ListItemPicker.d.ts
anthony9187: @soumyaloka that was my initial guess too, but I wanted to make sure. @ImCoolNowRight yeah im just concerned with making sure that the TS reflects the JS, do have a place where the object shape is defined?

ImCoolNowRight: It's a whole Marquee component. The Row is just a bunch of Marquees as items, and the `selected` getter is just returning whatever Marquee we're on.

PR #294, opened 2023-07-25
fix(Slider): update TS structure

soumyaloka:
## Description

This PR updates the Slider declarations 

## References

LUI-957
LUI-930

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Slider/SliderLarge.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/Slider/SliderLarge.d.ts
@@ -17,7 +17,7 @@
  */
 
 import lng from '@lightningjs/core';
-import Slider from '.';
+import Slider from './Slider';
 
 declare class SliderLarge<
   TemplateSpec extends Slider.TemplateSpec = Slider.TemplateSpec,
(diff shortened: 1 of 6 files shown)


anthony9187 (commented):
this looks great, two things:

1. let's move the `SliderStyle` export in Slider.d.ts to the bottom so we're exporting everything from one line
2. The SliderLarge.d.ts file needs to be updated as well

ImCoolNowRight (approved):

packages/@lightningjs/ui-components/src/components/Slider/SliderLarge.d.ts
```diff
@@ -16,6 +16,12 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import lng from '@lightningjs/core';
 import Slider from '.';
```
anthony9187: can we update this import line to `./Slider` so we're not importing from the index file?

soumyaloka: yep, I was little confused but yeah ./Slider will be explicit. Updated, Thanks!

anthony9187 (approved):

PR #300, opened 2023-07-26
chore: update alpha to theme value

salhan12:
## Description

Audited and updated alpha values to theme value across the components. 

## References

LUI-920


ImCoolNowRight:
Since you've updated the alpha values in a number of places, could you also do the below?

- ListItem.styles.js has an `alpha` property
- Slider.styles.js has an `arrowAlphaValue`
- Slider.styles.js also has an `arrowAlphaValueLimit`. That might be worth asking the team or design if there is a token it should be tied to.

erautenberg (approved):

soumyaloka (commented):
LGTM, Thanks for updating!

soumyaloka (approved):

PR #307, opened 2023-07-27
fix(Icon): bug in Icon update

joshhowenstine:
## Description

Fixes issue in previous Icon component updates causing colors not to update in button correctly. 

## References

[LUI-940](https://ccp.sys.comcast.net/browse/LUI-940) - This is a two part fix. The icon shift in button will be handled in another PR.

## Testing

Test all components that use Icon. 

Button should change icon color when has focus


packages/@lightningjs/ui-components/src/components/Icon/Icon.js
```diff
@@ -92,15 +77,17 @@ export default class Icon extends Base {
...
+    const { icon, w, h } = this;
+    const template = { w, h };
+    template.texture = {
+      type: CustomImageTexture,
+      w,
+      h,
+      src: icon
+    };
+    template.color = getValidColor(this.color || this.style.color);
+    return template;
```
erautenberg: can this whole function be simplified to something like this?
```js
_getIconPatch() {
  return {
    w: this.w,
    h: this.h,
    texture: {
      type: CustomImageTexture,
      w: this.w,
      h: this.h,
      src: this.icon
    },
    color: getValidColor(this.color || this.style.color)
  };
}
```

packages/@lightningjs/ui-components/src/components/Icon/Icon.js
```diff
@@ -64,25 +64,10 @@ export default class Icon extends Base {
...
-          color: getValidColor(this.style.color)
-        };
-      }
-    } else if (this.style.color) {
-      this.color = getValidColor(this.style.color);
-    }
 
-    // setting the radius on the Icon component
+    const template = this._getIconTemplate();
+    this.patch(template);
```
erautenberg: if the templates variable isn't being used anywhere else, maybe put it in one line? `this.patch(this _getIconTemplate());`

packages/@lightningjs/ui-components/src/components/Icon/Icon.js
erautenberg: Also might want to only apply the color if the color is passed in. I think this might break things for SVGs, just double check those

anthony9187: i think the root object would also need w and h? but yeah i think this is a easier to follow too

erautenberg: @anthony9187 you're right, just missed it in the first pass, updated the above for clarity

joshhowenstine: Should be handled now. Good callout

packages/@lightningjs/ui-components/src/components/Icon/Icon.js
```diff
@@ -20,87 +20,156 @@ import lng from '@lightningjs/core';
...
 import * as styles from './Icon.styles.js';
 import { context } from '../../globals';
-import { stringifyCompare, getValidColor } from '../../utils';
+import { getValidColor } from '../../utils';
 import CustomImageTexture from '../../textures/CustomImageTexture';
 
+/**
+ * Icon component that displays different types of icons, supporting SVGs and images.
+ * @extends Base
+ */
```
anthony9187, line 29: is the plan to start including these jsdoc blocks for methods in all classes? i dont have a problem with them, just asking from a code standards/style guide standpoint

joshhowenstine: I would like to, but yes good callout. We probably need to decide as a team what the standards are.

anthony9187: yeah - i dont want to hold up this PR so it's fine, but we should have a conversation around how we'd want to do this, and also make sure we aren't conflicting with the TS files because I'm pretty sure VSCode will pick these up

erautenberg (commented):
Not sure if it's related, but the icon seems fine in Button, but overlaps in Control:
http://localhost:8000/?path=/story/components-control--control&args=icon:lightning

packages/@lightningjs/ui-components/src/components/Icon/Icon.js
joshhowenstine: The potential TS collision is an extremely good point @anthony9187

joshhowenstine:
> Not sure if it's related, but the icon seems fine in Button, but overlaps in Control:
> http://localhost:8000/?path=/story/components-control--control&args=icon:lightning

I'm not sure exactly what the issue is at this point. However, I believe this is an issue that has surfaced in Control. If it starts with an icon set it works fine. Ok if we make a follow up ticket?

erautenberg:
> > Not sure if it's related, but the icon seems fine in Button, but overlaps in Control:
> > http://localhost:8000/?path=/story/components-control--control&args=icon:lightning
> 
> I'm not sure exactly what the issue is at this point. However, I believe this is an issue that has surfaced in Control. If it starts with an icon set it works fine. Ok if we make a follow up ticket?

That's fine, can you coordinate a ticket with Nassar please?

erautenberg (approved):

ImCoolNowRight (approved):
Looks good!

PR #311, opened 2023-07-28
fix(label): adds type declarations

arwehrman:
## Description

This PR updates TS type declarations for Label component

## References

LUI-955

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Label/Label.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/Label/Label.d.ts
@@ -18,9 +18,9 @@
 
 import lng from '@lightningjs/core';
 import Base from '../Base';
-import type { Color, StylePartial } from '../../types/lui';
+import { Color, StylePartial } from '../../types/lui';
 
-export type LabelStyle = {
+type LabelStyle = {
   backgroundColor: Color;
   paddingX: number;
   paddingY: number;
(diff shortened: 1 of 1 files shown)


soumyaloka (approved):

packages/@lightningjs/ui-components/src/components/Label/Label.d.ts
```diff
@@ -18,17 +18,36 @@
...
+  }
+  export interface TypeConfig extends lng.Component.TypeConfig {
+    SignalMapType: SignalMap;
+  }
+  export type SignalMap = {
+    updateBackground(): void;
+  };
+}
+declare class Label<
+  TemplateSpec extends Label.TemplateSpec = Label.TemplateSpec
```
soumyaloka: May be we should add TypeConfig here like
` TypeConfig extends Label.TypeConfig = Label.TypeConfig`
But I would like to defer to @anthony9187,  if that is needed all the time or not.

anthony9187: yup, we need to pass the TypeConfig to the class declaration in order for the SignalMapType to get properly attached to the class. In general we should always be including a TypeConfig even if its just
```
```

anthony9187 (approved):

PR #312, opened 2023-07-29
feat: yarn zero install

joshhowenstine:


mhughesacn:
Hi Josh: I know that this is WIP at the moment but heads up that this pull req is a real problem for CMF in its present form: it contains cached material.  There are two problems with this: first, the changes are too big to scan, containing huge amounts of external open sourced material.  The best practice is to fetch, not store.  Second, while we have sometimes stored external OSS after approval from RDK, the quantity here is not manageable: the .yarn directory is 124MB, far in excess of most repositories.  It increases the size of the repo by an order of magnitude.  If this is intended to be permanent, we will have to discuss.
Martin Hughes, RDK CMF Compliance Team

joshhowenstine:
Not required at this point

PR #314, opened 2023-07-31
fix(Control): update TS structure

soumyaloka:
## Description

In this PR updated typescript declarations file to latest TS template structure for ButtonSmall and Control components

## References

LUI-837

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Button/ButtonSmall.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/Button/ButtonSmall.d.ts
@@ -25,10 +25,6 @@ type ButtonSmallStyle = ButtonStyle & {
   suffixH: number;
 };
 
-declare namespace ButtonSmall {
-  export type TemplateSpec = Button.TemplateSpec;
-}
-
 declare class ButtonSmall<
   TemplateSpec extends Button.TemplateSpec = Button.TemplateSpec,
   TypeConfig extends lng.Component.TypeConfig = lng.Component.TypeConfig
--- a/packages/@lightningjs/ui-components/src/components/Button/ButtonSmall.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/Button/ButtonSmall.d.ts
@@ -19,7 +19,9 @@
 import lng from '@lightningjs/core';
 import Button from './Button';
 
-export default class ButtonSmall<
-  TemplateSpec extends Button.TemplateSpec,
-  TypeConfig extends lng.Component.TypeConfig
+declare class ButtonSmall<
+  TemplateSpec extends Button.TemplateSpec = Button.TemplateSpec,
+  TypeConfig extends lng.Component.TypeConfig = lng.Component.TypeConfig
 > extends Button<TemplateSpec, TypeConfig> {}
+
+export default ButtonSmall;
--- a/packages/@lightningjs/ui-components/src/components/Button/ButtonSmall.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/Button/ButtonSmall.d.ts
@@ -19,19 +19,16 @@
 import lng from '@lightningjs/core';
 import Button, { ButtonStyle } from './Button';
 import { StylePartial } from '../../types/lui';
-import type { TextBoxStyle } from '../TextBox';
 
 type ButtonSmallStyle = ButtonStyle & {
-  minWidth: number;
-  paddingX: number;
   prefixH: number;
   suffixH: number;
-  textStyle: TextBoxStyle;
 };
 
 declare namespace ButtonSmall {
   export type TemplateSpec = Button.TemplateSpec;
 }
+
 declare class ButtonSmall<
   TemplateSpec extends Button.TemplateSpec = Button.TemplateSpec,
   TypeConfig extends lng.Component.TypeConfig = lng.Component.TypeConfig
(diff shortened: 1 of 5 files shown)


arwehrman (commented):
ControlSmall needs to be updated as well in this PR

packages/@lightningjs/ui-components/src/components/Button/ButtonSmall.d.ts
```diff
@@ -17,9 +17,27 @@
...
+  prefixH: number;
+  suffixH: number;
+  textStyle: TextBoxStyle;
+};
+
+declare namespace ButtonSmall {
+  export type TemplateSpec = Button.TemplateSpec;
+}
+declare class ButtonSmall<
+  TemplateSpec extends Button.TemplateSpec = Button.TemplateSpec,
```
ImCoolNowRight, line 29: These should be `ButtonSmall.TemplateSpec`

soumyaloka: Thanks for reviewing I will update this, I made some changes and removed wip label but looks like I forgot to push the changes.

soumyaloka: @ImCoolNowRight , Regarding your comment, I think that is correct because If there are no additional properties specific to ButtonSmall that need to be defined in its TemplateSpec so I think we can use Button.TemplateSpec, that way we are using the TemplateSpec structure defined in Button for ButtonSmall.

packages/@lightningjs/ui-components/src/components/Control/ControlSmall.styles.js
```diff
@@ -34,7 +34,6 @@ export const base = theme => {
     },
     prefixH: theme.spacer.md * 5,
     radius,
-    textStyle: theme.typography.button2,
```
soumyaloka, line 37: Deleted this because it is already defined in ButtonSmall component(Control extends ButtonSmall).

packages/@lightningjs/ui-components/src/components/Button/ButtonSmall.d.ts
arwehrman: So think you need to extend the TemplateSpec that you defined in the namespace then extend Button
so 
`declare class ButtonSmall<
  TemplateSpec extends ButtonSmall.TemplateSpec = ButtonSmall.TemplateSpec,
> extends Button<TemplateSpec, TypeConfig> {`

packages/@lightningjs/ui-components/src/components/Button/ButtonSmall.d.ts
```diff
@@ -17,9 +17,24 @@
...
-  TypeConfig extends lng.Component.TypeConfig
-> extends Button<TemplateSpec, TypeConfig> {}
+type ButtonSmallStyle = ButtonStyle & {
+  prefixH: number;
+  suffixH: number;
+};
+
+declare namespace ButtonSmall {
+  export type TemplateSpec = Button.TemplateSpec;
+}
```
anthony9187: we can omit this

packages/@lightningjs/ui-components/src/components/Button/ButtonSmall.d.ts
anthony9187: as it is is correct but we shouldnt be declaring empty namespaces/interfaces, per my comment above

packages/@lightningjs/ui-components/src/components/Control/ControlSmall.styles.js
anthony9187: we'd still need it because otherwise it's not included in the returned object right?

packages/@lightningjs/ui-components/src/components/Control/Control.d.ts
```diff
@@ -17,26 +17,63 @@
...
+    /**
+     * flag that if `true`, hides the title when `Control` is in unfocused or disabled mode
+     */
+    shouldCollapse?: boolean;
+  }
+}
+
+declare class Control<
+  TemplateSpec extends Control.TemplateSpec = Control.TemplateSpec,
+  TypeConfig extends lng.Component.TypeConfig = lng.Component.TypeConfig
```
ImCoolNowRight, line 57: @anthony9187 Still trying to make sure I understand this correctly. Should this TypeConfig be extending the Button's TypeConfig instead, or is this correct?

packages/@lightningjs/ui-components/src/components/Control/Control.d.ts
```diff
@@ -17,26 +17,63 @@
...
-  h: number;
+type ControlStyle = ButtonSmallStyle & {
   iconStyle: Record<string, unknown>;
   logoStyle: LogoStyleObject;
   radius: lng.Tools.CornerRadius;
 };
 
-export default class Control extends ButtonSmall {
+declare namespace Control {
+  // TODO: need to replace Button with ButtonSmall here , but throws error. Is this correct now?
```
soumyaloka: @anthony9187 Need your eyes on this comment

anthony9187: is this an old comment from when we were extending button instead?

packages/@lightningjs/ui-components/src/components/Control/Control.d.ts
anthony9187: this is correct since Button doesn't have a TypeConfig and itself uses lng.Component's TypeConfig

anthony9187 (approved):

arwehrman (approved):

PR #315, opened 2023-07-31
fix(Gradient): update TS structure

soumyaloka:
## Description

In this PR , update typescript declarations to latest template structure

## References

LUI-945

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Gradient/Gradient.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/Gradient/Gradient.d.ts
@@ -25,10 +25,13 @@ type GradientStyle = {
   gradientTop: string;
   radius: lng.Tools.CornerRadius;
 };
-
+declare namespace Gradient {
+  export type TemplateSpec = Base.TemplateSpec;
+}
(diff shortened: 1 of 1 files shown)


arwehrman (approved):

packages/@lightningjs/ui-components/src/components/Gradient/Gradient.d.ts
```diff
@@ -18,15 +18,22 @@
...
+type GradientStyle = {
   gradientColor: Color;
+  gradientTop: string;
   radius: lng.Tools.CornerRadius;
 };
-
-export default class Gradient extends Base {
+declare namespace Gradient {
+  export type TemplateSpec = Base.TemplateSpec;
+}
```
anthony9187: we can omit this and just use `Base.TemplateSpec` below if there's no additional properties

soumyaloka: Ok, Got it. I will update that. Thanks!

anthony9187 (approved):

PR #316, opened 2023-07-31
fix(ProgressBar): rounded corners

slhay28:
## Description

Renders the progress within the bar to consistently have rounded corners to the same degree regardless of the progress width.

## References

LUI-869

## Testing

Move the progress bar progress higher and lower to see the consistent rounded edges.

## Automation
None

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/ProgressBar/ProgressBar.js
+++ b/packages/@lightningjs/ui-components/src/components/ProgressBar/ProgressBar.js
@@ -59,8 +59,8 @@ export default class ProgressBar extends Base {
   }
 
   _updateTextures() {
-
-    const w = this._getProgressWidth();
+    const p = this.w * this._progress;
+    const w = p <= 0 ? 0 : Math.min(p, this.w);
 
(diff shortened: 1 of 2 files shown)


arwehrman (approved):

ImCoolNowRight (commented):
I think the other part necessary for this is removing the `w` logic from `_updateProgress` on line 91so that the texture's width doesn't get adjusted any further. So `_updateProgress` will only handle updating the `alpha`.

packages/@lightningjs/ui-components/src/components/ProgressBar/ProgressBar.js
```diff
@@ -59,6 +59,9 @@ export default class ProgressBar extends Base {
   }
 
   _updateTextures() {
+    const p = this.w * this._progress;
+    const w = p <= 0 ? 0 : Math.min(p, this.w);
```
erautenberg: Since these same two lines are used farther down in the updateProgress method, how about we make this it's own method, maybe an assessor like `get _progressWidth`?

ImCoolNowRight (approved):

PR #319, opened 2023-08-01
feat(TextBox): add support for Marquee properties in TextBox

soumyaloka:
## Description

This PR adds support for the Marquee component in the TextBox. Now, when the 'marquee' prop is set in the TextBox, it will render the Marquee component with properties to control its behavior, such as 'repeat'.

## References

LUI-814

## Testing

1. Navigate to any component that uses the TextBox type.
2. Try adding the following properties to the TextBox:
```
type: TextBox,
marquee: true,
marqueeProps: {
repeat: 2
},
```
3. Verify that the TextBox content correctly scrolls based on the specified repeat time.

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/TextBox/TextBox.js
+++ b/packages/@lightningjs/ui-components/src/components/TextBox/TextBox.js
@@ -61,7 +61,13 @@ export default class TextBox extends Base {
   }
 
   static get properties() {
-    return ['content', 'fixed', 'marquee', ...InlineContent.properties];
+    return [
+      'content',
+      'fixed',
+      'marquee',
+      ...InlineContent.properties,
+      'marqueProps'
+    ];
   }
 
   _setDimensions(w, h) {
@@ -156,11 +162,6 @@ export default class TextBox extends Base {
     if (this._textStyleSet.maxLinesSuffix) {
       inlineContentPatch.maxLinesSuffix = this._textStyleSet.maxLinesSuffix;
     }
-    if (this._textStyleSet.textAlign) {
-      inlineContentPatch.justify = utils.convertTextAlignToFlexJustify(
-        this._textStyleSet.textAlign
-      );
-    }
 
     this.patch({
       alpha: 1,
@@ -233,6 +234,7 @@ export default class TextBox extends Base {
         h: this.h,
         y: this.style.offsetY,
         x: this.style.offsetX,
+        ...this._marqueProps,
         signals: {
           marqueeContentLoaded: '_loadedMarqueeContent'
         }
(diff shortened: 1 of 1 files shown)


packages/@lightningjs/ui-components/src/components/TextBox/TextBox.js
```diff
@@ -61,7 +61,13 @@ export default class TextBox extends Base {
   }
 
   static get properties() {
-    return ['content', 'fixed', 'marquee', ...InlineContent.properties];
+    return [
+      'content',
+      'fixed',
+      'marquee',
+      ...InlineContent.properties,
```
anthony9187, line 68: we should list the destructured item first, unless there's a specific reason not to(a property we explicitly we want overridden or something)

ImCoolNowRight:
Can you provide some more detailed testing instructions? I'm having trouble getting this working with components using TextBox, and I'm not sure if I'm just testing it incorrectly.

packages/@lightningjs/ui-components/src/components/TextBox/TextBox.js
anthony9187: lets list destructures as the first item in the array

packages/@lightningjs/ui-components/src/components/TextBox/TextBox.js
```diff
@@ -233,6 +234,7 @@ export default class TextBox extends Base {
         h: this.h,
         y: this.style.offsetY,
         x: this.style.offsetX,
+        ...this.marqueeProps,
```
anthony9187: lets move this destructure to the top of the object so we dont accidentally override any declared properties

packages/@lightningjs/ui-components/src/components/TextBox/TextBox.js
```diff
@@ -271,6 +273,14 @@ export default class TextBox extends Base {
     }
   }
 
+  get marqueeProps() {
+    return this._marqueeProps ?? {};
+  }
+
+  set marqueeProps(props) {
+    this._marqueeProps = props;
+  }
```
anthony9187: since these are in the `properties` array we should use the `_getMarqueeProps` and `_setMarqueeProps` functions so we aren't overriding the accessors created by `withUpdates`

packages/@lightningjs/ui-components/src/components/TextBox/TextBox.js
```diff
@@ -271,6 +273,14 @@ export default class TextBox extends Base {
     }
   }
 
+  _getMarqueeProps() {
+    return this._marqueeProps ?? {};
+  }
+
+  _setMarqueeProps(props) {
+    this._marqueeProps = props;
```
anthony9187: if all this isn't mutating the value or handling a side effect, it can be omitted. the way the `_set<PropName>` methods work is they get called at the beginning of the setter method, process the value, and then pass it back to the setter. One added benefit is we don't need matching pairs like we do if we were overriding the accessor methods

ImCoolNowRight (approved):

anthony9187 (approved):

PR #320, opened 2023-08-01
fix(ProgressBar): update TS structure

soumyaloka:
## Description

This PR updates typescript declaration to new template structure.

## References

LUI-962

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/ProgressBar/ProgressBar.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/ProgressBar/ProgressBar.d.ts
@@ -18,7 +18,7 @@
 
 import lng from '@lightningjs/core';
 import Base from '../Base';
-import type { Color, StylePartial } from '../../types/lui';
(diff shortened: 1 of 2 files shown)


packages/@lightningjs/ui-components/src/components/ProgressBar/ProgressBar.d.ts
```diff
@@ -18,23 +18,38 @@
...
   progressColor: Color;
   radius: lng.Tools.CornerRadius;
 };
 
-export default class ProgressBar extends Base {
-  progress: string;
+declare namespace ProgressBar {
+  export interface TemplateSpec extends Base.TemplateSpec {
+    /**
+     * percentage of the current progress from 0 to 1
```
anthony9187: I know this was copied over from the mdx but it doesn't quite make sense - we're passing a decimal value, not a percentage

anthony9187 (approved):

ImCoolNowRight (approved):

PR #323, opened 2023-08-02
feat(Key): update TS structure

soumyaloka:
## Description

In this PR updated Key typescript declarations file to latest TS template structure

## References

LUI-951

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Key/Key.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/Key/Key.d.ts
@@ -56,8 +56,6 @@ declare namespace Key {
      * If true, pressing the key will trigger the $toggleKeyboard event. If false, the key will trigger the $onSoftKey event.
      */
     toggle?: boolean;
-
-    style?: StylePartial<KeyStyle>;
   }
 }
 declare class Key<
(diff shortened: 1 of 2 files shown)


arwehrman (approved):

ImCoolNowRight (approved):

packages/@lightningjs/ui-components/src/components/Key/Key.d.ts
```diff
@@ -27,17 +28,85 @@ type SizeObject = {
...
+
+    style?: StylePartial<KeyStyle>;
+  }
+  export interface TypeConfig extends lng.Component.TypeConfig {
+    SignalMapType: SignalMap;
+  }
+  /**
+   * Signal map interface for Key component
+   */
+  export type SignalMap = {
```
anthony9187: It doesn't seem like Key has any signals?

edit: yeah these are fireAncestors calls. we can omit these for now, I'm prototyping out a way to include an app-level augment

packages/@lightningjs/ui-components/src/components/Key/Key.d.ts
```diff
@@ -27,17 +28,85 @@ type SizeObject = {
...
+   */
+  onEnter?: (key: string, toggle: boolean) => void;
+
+  /**
+   * Logo to be placed to the left of the title and description
+   */
+  size?: string;
+
+  /**
+   * used to toggle `$toggleKeyboard`
```
anthony9187: can we add what true/false do here?

packages/@lightningjs/ui-components/src/components/Key/Key.d.ts
soumyaloka: Sure, I will remove that now then, Thanks!

PR #324, opened 2023-08-02
fix(TextBox): updates TS structure

arwehrman:
## Description

This PR updates TextBox and Inline type declarations

## References

LUI-949 

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.d.ts
@@ -133,7 +133,6 @@ declare class InlineContent<
 > extends Base<TemplateSpec, TypeConfig> {
   /**
    * list of data to render inline
-   * TODO: Feel like there should be more to this
    */
   content?: TextContent[];
   /**
--- a/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.d.ts
@@ -20,6 +20,7 @@ import lng from '@lightningjs/core';
 import Badge from '../Badge';
 import Base from '../Base';
 import { StylePartial } from '../../types/lui';
+import TextContent from '../TextBox/TextBox';
 
 type JustifyContent =
   | 'flex-start'
@@ -28,16 +29,6 @@ type JustifyContent =
   | 'space-between'
   | 'space-around'
   | 'space-evenly';
-
-type ContentBase = {
-  title?: string;
-  announce?: string;
-  style?: string | Record<string, unknown>;
-};
-
-export type TextContent =
-  | string
-  | (ContentBase & ({ text: string } | { icon: string } | { badge: string }));
 
 type FlexItem = {
   grow?: number;
--- a/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.d.ts
@@ -161,9 +161,9 @@ declare class InlineContent<
    */
   contentWrap?: boolean;
   /**
-   * set of nested objects containing [Lightning Text](https://rdkcentral.github.io/Lightning/docs/textures/text) properties which can be referenced in the [Content Element Properties](#content-element-properties) to apply styling to individual text elements
+   * set of nested objects containing [Lightning Text](https://lightningjs.io/api/lightning-core/interfaces/lightning.textures.texttexture.settings) properties which can be referenced in the [Content Element Properties](#content-element-properties) to apply styling to individual text elements
    */
-  customStyleMappings?: Record<string, unknown>;
+  customStyleMappings?: Record<string, lng.textures.TextTexture.Settings>;
   /**
    * maximum number of lines to render before truncation
    */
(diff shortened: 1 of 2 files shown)


soumyaloka (approved):
Looks great!

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.d.ts
```diff
@@ -77,14 +77,106 @@ export type InlineContentStyle = {
...
+  export interface TypeConfig extends lng.Component.TypeConfig {
+    SignalMapType: SignalMap;
+  }
+  export type SignalMap = {
+    // signals: _createIcon  & _createBadge
+
+    _createIcon(base, iconProps);
+  };
+
+  // events: $loadedInlineContent(): void;
```
anthony9187: the only signal should be `loadedInlineContent`, the others are part of child components

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.d.ts
```diff
@@ -77,14 +77,106 @@ export type InlineContentStyle = {
...
+   */
   justify?: JustifyContent;
+  /**
+   * determines whether the containing flexbox should wrap the content onto the next line
+   */
   contentWrap?: boolean;
+  /**
+   * set of nested objects containing [Lightning Text](https://rdkcentral.github.io/Lightning/docs/textures/text) properties which can be referenced in the [Content Element Properties](#content-element-properties) to apply styling to individual text elements
+   */
   customStyleMappings?: Record<string, unknown>;
```
anthony9187: I think we can update this `unknown` to be the text properties from lightning/core

packages/@lightningjs/ui-components/src/components/TextBox/TextBox.d.ts
```diff
@@ -18,27 +18,92 @@
 
 import lng from '@lightningjs/core';
 import Base from '../Base';
-import type { StylePartial } from '../../types/lui';
-import type { TextContent } from '../InlineContent/InlineContent';
+import { StylePartial } from '../../types/lui';
+import TextContent from '../InlineContent';
```
anthony9187: I don't think InlineContent exports this?

packages/@lightningjs/ui-components/src/components/TextBox/TextBox.d.ts
```diff
@@ -18,27 +18,92 @@
...
+    marquee?: boolean;
+  }
+  export interface TypeConfig extends lng.Component.TypeConfig {
+    SignalMapType: SignalMap;
+  }
+
+  export type SignalMap = {
+    /** emits when marquee content has been patched on update
+     */
+    _loadedMarqueeContent(): void;
```
anthony9187: the signals should be `textBoxChanged` and `willMarquee`

packages/@lightningjs/ui-components/src/components/TextBox/TextBox.d.ts
arwehrman: I updated TextContent to match the current pattern we have for other components

anthony9187: can we also move the TextContent export to the bottom of the file with the other exports? I'd prefer to group all of our exports together in one place

anthony9187: also part of this we now have TextContent declared in both InlineContent and TextBox, can we remove TextContent from InlineContent and import it from TextBox?(I _think_ thats how it should be setup)

anthony9187:
we should wait to merge this until https://github.com/rdkcentral/Lightning-UI-Components/pull/319 goes in so we can include the new `marqueeProps` property in the declaration file

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.d.ts
```diff
@@ -77,14 +68,101 @@ export type InlineContentStyle = {
...
+     */
+    justify?: JustifyContent;
+    /**
+     * determines whether the containing flexbox should wrap the content onto the next line
+     */
+    contentWrap?: boolean;
+    /**
+     * set of nested objects containing [Lightning Text](https://rdkcentral.github.io/Lightning/docs/textures/text) properties which can be referenced in the [Content Element Properties](#content-element-properties) to apply styling to individual text elements
+     */
+    customStyleMappings?: Record<string, unknown>;
```
anthony9187: I don't believe `Record<string, unknown>` will allow for nested objects. if the expected object is something like:
```
{
  elementName: {
    customStyle: value,
    customStyle2: value
  }
}
```

something like this may work better:
```
Record<string, Record<string, unknown>>
```

 if there would be more nesting beyond that, then:
```
Record<string, Record<string, any>>
``` 
as much as I'd prefer to avoid any

edit: or perhaps this just needs to match line 148 below?

packages/@lightningjs/ui-components/src/components/TextBox/TextBox.d.ts
```diff
@@ -18,27 +18,107 @@
...
+import { StylePartial } from '../../types/lui';
+import Marquee from '../Marquee';
+import InlineContent from '../InlineContent';
 
-export type TextBoxStyle = {
+// TODO: Properly export TextContent from root level of ui-components so we don't have to grab it from InlineContent
+type ContentBase = {
+  title?: string;
+  announce?: string;
+  style?: string | Record<string, unknown>;
```
anthony9187: is this object expected to be the same as the `textStyle` in the TextBoxStyle type below? we could use that same lng.textures.TextTexture.Settings here as well if so

anthony9187 (approved):

PR #326, opened 2023-08-03
feat(Input): update TS structure

soumyaloka:
## Description

This PR updates Input type declarations

## References

LUI-953

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Input/Input.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/Input/Input.d.ts
@@ -21,7 +21,7 @@ import Button, { ButtonStyle } from '../Button';
 import { TextBoxStyle } from '../TextBox';
 import { StylePartial } from '../../types/lui';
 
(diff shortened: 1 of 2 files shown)


arwehrman (approved):

packages/@lightningjs/ui-components/src/components/Input/Input.d.ts
```diff
@@ -30,30 +30,129 @@ export type InputStyle = ButtonStyle & {
...
-  get _HelpText(): lng.Component;
-  get _Cursor(): lng.Component;
-  get _HiddenContent(): lng.Component;
+  // methods
 
-  // public methods
+  /**
+   * clears the content of the input container
+   */
   clear: () => void;
```
anthony9187: all these class methods should be written as 
```
methodName(v: parameterType): returnValue;
```
so this one would be
```
clear(): void;
```

ImCoolNowRight (approved):

PR #328, opened 2023-08-03
feat(Knob): update TS structure

soumyaloka:
## Description

This PR updates Knob type declarations

## References

LUI-954

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Knob/Knob.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/Knob/Knob.d.ts
@@ -18,17 +18,19 @@
 
 import lng from '@lightningjs/core';
 import Base from '../Base';
-import type { Color, StylePartial } from '../../types/lui';
+import { Color, StylePartial } from '../../types/lui';
 
-export type KnobStyle = {
+type KnobStyle = {
   circleColor: Color;
   radius: lng.Tools.CornerRadius;
 };
 
-export default class Knob extends Base {
+declare class Knob<
+  TemplateSpec extends Base.TemplateSpec = Base.TemplateSpec,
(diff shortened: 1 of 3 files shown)


arwehrman (approved):

packages/@lightningjs/ui-components/src/components/Knob/Knob.d.ts
```diff
@@ -18,17 +18,19 @@
 
 import lng from '@lightningjs/core';
 import Base from '../Base';
-import type { Color, StylePartial } from '../../types/lui';
+import { Color, StylePartial } from '../../types/lui';
 
-export type KnobStyle = {
+type KnobStyle = {
   circleColor: Color;
```
anthony9187, line 24: is this right? circleColor only appears in tones/modes, so would it be accessible via `Knob.style.circleColor`?

soumyaloka: Yes I think circleColor property is accessible via Knob.style.circleColor, it is used in slider component.

packages/@lightningjs/ui-components/src/components/Knob/Knob.d.ts
```diff
@@ -18,17 +18,19 @@
...
   radius: lng.Tools.CornerRadius;
 };
```
arwehrman, line 27: I think you still need to add the namespace and then update the class with the namespace but should check with Anthony
```
declare namespace Knob {
  export type TemplateSpec = Base.TemplateSpec;
}
```

anthony9187: nah in cases where we aren't making changes to the TemplateSpec or TypeConfig we should omit them and just pass the parent's interfaces

anthony9187 (approved):

PR #333, opened 2023-08-07
feat(CardContent): update TS structure

soumyaloka:
## Description

This PR updates CardContent TS declaration and removed duplicate styles which are already defined in parent components.

## References

LUI-833

## Checklist


packages/@lightningjs/ui-components/src/components/CardContent/CardContentHorizontal.d.ts
```diff
@@ -16,8 +16,15 @@
...
-import CardContent, { CardContentStyle } from './CardContent';
+import lng from '@lightningjs/core';
+import CardContent from './CardContent';
 
-export type CardContentHorizontalStyle = CardContentStyle;
+declare namespace CardContentHorizontal {
+  export type TemplateSpec = CardContent.TemplateSpec;
+}
+declare class CardContentHorizontal<
+  TemplateSpec extends CardContent.TemplateSpec = CardContent.TemplateSpec,
```
arwehrman, line 23: I believe this should be `CardContentHorizontal.TemplateSpec = CardContentHorizontal.TemplateSpec`
 the rest is correct

packages/@lightningjs/ui-components/src/components/CardContent/CardContentHorizontalLarge.d.ts
```diff
@@ -16,10 +16,12 @@
...
 
-import CardContentHorizontal, {
-  CardContentHorizontalStyle
-} from './CardContentHorizontal';
+import lng from '@lightningjs/core';
+import CardContentHorizontal from './CardContentHorizontal';
 
-export type CardContentHorizontalLargeStyle = CardContentHorizontalStyle;
+declare class CardContentHorizontalLarge<
+  TemplateSpec extends CardContentHorizontal.TemplateSpec = CardContentHorizontal.TemplateSpec,
```
arwehrman: Same here: `CardContentHorizontalLarge.TemplateSpec = CardContentHorizontalLarge.TemplateSpec`

arwehrman:
For the class declarations:

Update the class declaration to include the namespace's types 

`declare class Surface<TemplateSpec extends Surface.TemplateSpec = Surface.TemplateSpec, TypeConfig extends lng.Component.TypeConfig = lng.Component.TypeConfig> extends Base<TemplateSpec, TypeConfig>`

in the above line, we're making Surface a generic by adding the `<>` and then passing the above TemplateSpec and re-defining the default TypeConfig using `TypeConfig extends lng.Component.TypeConfig` Those are then passed to Base in `Base<TemplateSpec, TypeConfig>`

packages/@lightningjs/ui-components/src/components/CardContent/CardContentHorizontal.d.ts
soumyaloka: Thanks! My thought process was like since CardContentHorizontal extends CardContent and doesn't have any additional properties or styles of its own, using the TemplateSpec from CardContent makes sense here.
May be I am wrong.

anthony9187: that's correct, but in that case we should be omitting the namespace entirely instead of declaring an empty TemplateSpec

anthony9187:
any place that an empty TypeConfig is being declared, we should remove it entirely. We don't need to re-declare emtpy interfaces just for the sake of passing a renamed interface. Just pass the parent's interfaces

packages/@lightningjs/ui-components/src/components/CardContent/CardContentHorizontal.d.ts
soumyaloka: But since CardContentHorizontalLarge extends CardContentHorizontal, having an empty namespace will throw an error, right?

anthony9187: nope, it would just reference the parent's interfaces

packages/@lightningjs/ui-components/src/components/CardContent/CardContent.d.ts
```diff
@@ -17,31 +17,87 @@
...
+declare namespace CardContent {
+  export interface TemplateSpec extends Card.TemplateSpec {
+    /**
+     * when collapseToMetadata is true and shouldCollapse is enabled, the image collapses to show the metadata
+     */
+    collapseToMetadata?: boolean;
+    /**
+     * object containing all properties supported in the MetadataCardContent component
+     */
+    metadata?: object;
```
anthony9187, line 48: this should be updated to a MetadataCardContent patch object once it's d.ts has been updated

packages/@lightningjs/ui-components/src/components/CardContent/CardContent.d.ts
```diff
@@ -17,31 +17,87 @@
...
+     */
+    shouldCollapse?: boolean;
+    /**
+     *  path to artwork image
+     */
+    src: string;
+    /**
+     * object containing all properties supported in the Tile component
+     */
+    tile?: object;
```
anthony9187, line 65: same here, this should be a Tile patch object once Tile's updated

anthony9187 (approved):

arwehrman (approved):

PR #334, opened 2023-08-07
fix(Radio): update TS structure

slhay28:
## Description

updates TS structure in radio and radioSmall

## References

LUI-964

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Radio/Radio.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/Radio/Radio.d.ts
@@ -50,6 +50,12 @@ declare class Radio<
    */
   checked?: boolean;
 
+    // Method
+  /**
+   * Toggles the radio state and updates the UI.
+   */
+  toggle(): void;
+
   get style(): RadioStyle;
   set style(v: StylePartial<RadioStyle>);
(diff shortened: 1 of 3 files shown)


packages/@lightningjs/ui-components/src/components/Radio/Radio.d.ts
```diff
@@ -32,13 +32,43 @@ export type RadioStyle = {
...
+> extends Base<TemplateSpec, TypeConfig> {
+  /**
+   * status of the radio button
+   */
   checked?: boolean;
+
   get style(): RadioStyle;
   set style(v: StylePartial<RadioStyle>);
 
   // tags
```
soumyaloka, line 62: This looks great but it seems like we missed adding the 'toggle' method here.

soumyaloka: Looks like alpha style property is not added in docs file, it would be great if you can add that as well as part of this PR. Thank You!

anthony9187 (approved):
after @soumyaloka'a requested changes, this'll be good to merge

soumyaloka (approved):

PR #348, opened 2023-08-16
fix(ControlRow): update TS structure

soumyaloka:
## Description

This PR updates ControlRow TS structure 

## References

LUI-838


packages/@lightningjs/ui-components/src/components/ControlRow/ControlRow.d.ts
```diff
@@ -16,34 +16,97 @@
...
+> extends TitleRow<TemplateSpec, TypeConfig> {
+  /**
+   * additional space to be added between controls and content items
+   */
   extraItemSpacing?: number;
+
   get style(): ControlRowStyle;
   set style(v: StylePartial<ControlRowStyle>);
 
-  loadMoreItems(): void;
```
anthony9187, line 39: we should still have this method in the declaration, but I'm also a little confused as to the purpose of this method

packages/@lightningjs/ui-components/src/components/ControlRow/ControlRow.d.ts
```diff
@@ -16,34 +16,97 @@
...
-import type lng from '@lightningjs/core';
-import type { StylePartial } from '../../types/lui';
+import lng from '@lightningjs/core';
+import { StylePartial } from '../../types/lui';
 import TitleRow, { TitleRowStyle } from '../TitleRow';
 
-export type ControlRowStyle = TitleRowStyle & {
+type ControlRowItem = {
+  extraItemSpacing?: number;
+};
```
anthony9187: why was this changed? this isn't a valid component type, and it doesn't seem like `extraItemSpacing` is a valid property on either controls or items

packages/@lightningjs/ui-components/src/components/ControlRow/ControlRow.d.ts
```diff
@@ -60,30 +56,37 @@ declare class ControlRow<
...
+  addContentItems(items: unknown[]): void;
 
   /**
    * Adds items at a specified index of the row.
-   * @param items An array of ControlRowItem to be added.
+   * @param items An array of items to be added.
    * @param index The index where the items should be added.
    */
-  addContentItemsAt(items: ControlRowItem[], index: number): void;
+  addContentItemsAt(items: unknown[], index: number): void;
```
anthony9187: is there a reason we're using unknown here instead of a stricter type?

packages/@lightningjs/ui-components/src/components/ControlRow/ControlRow.d.ts
```diff
@@ -16,34 +16,105 @@
...
+/**
+ * The signal is emitted when an item is selected after an index determined by the lazyLoadBuffer property.
+ */
+export type SignalMap = {
+  loadMoreItems(): void;
+};
+
+declare class ControlRow<
+  TemplateSpec extends ControlRow.TemplateSpec = ControlRow.TemplateSpec,
+  TypeConfig extends lng.Component.TypeConfig = lng.Component.TypeConfig
```
anthony9187: this should extends FocusManager.TypeConfig the same way TitleRow does

anthony9187 (approved):

arwehrman (approved):

PR #352, opened 2023-08-22
fix(InlineContent): reset tracked x position after new line

THoj13:
## Description
This resolves an issue where the number of lines rendered by InlineContent sometimes exceeds the maxLines. The source of the issue was where we calculate which line each piece of text/icon/badge is rendered on. When a new line is in the content string, the next item should start on a new line. Prior to this change, that would increase the line number for the next item and set the ending x position of the last item rendered to the width of the InlineContent component. Both of those changes individually are meant to denote a new line should be started, so doing both resulted in adding 2 lines. This change still increases the line number but sets the tracked end x position of the last item to the beginning of the component.

## References
LUI-1016

## Testing
This issue was only reproducible in the inner source repo. 
1. Navigate to "InlineContent / With Inline Content String" story
2. set contentWrap to true and set maxLines to 4
ER: 4 lines of text are rendered

## Automation
Bug initially reported by Automation team.

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
+++ b/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
@@ -18,7 +18,7 @@
 
 import Icon from '../Icon';
 import Badge from '../Badge';
-import { parseInlineContent, flatten } from '../../utils';
+import { parseInlineContent, flatten, measureTextWidth } from '../../utils';
 import Base from '../Base';
 import * as styles from './InlineContent.styles.js';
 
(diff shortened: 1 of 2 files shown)


erautenberg:
Seems to work for maxLines of 5, but not 6, I see 7 lines displayed still.

Also, mind setting a default value in the story controls so we don't have to hit the "set number" button?

erautenberg:
Just did some experimenting. It actually seems like the "maxLinesSuffix" is what causes it to wrap to the next line, so maybe this isn't be included in the calculation for truncation?

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
```diff
@@ -197,7 +197,7 @@ export default class InlineContent extends Base {
 
       if (isNewLineElement) {
         line++;
-        contentEndX = w;
+        contentEndX = 0;
```
erautenberg, line 231: maybe this should be the width of the `maxLinesSuffix` ?

erautenberg (approved):

ImCoolNowRight (approved):

PR #357, opened 2023-08-24
feat(TextBox): allow defining whether or not texture should display after loading

THoj13:
## Description
Adds new property, `displayOnLoad`, which defines if the component should be visible after the text texture has loaded. To maintain current behavior, this property is defaulted to `true`.

**Related:** This property should be added to the updated TypeScript definition for this component once both are merged. https://github.com/rdkcentral/Lightning-UI-Components/pull/324

## Testing
The TextBox component stories should be unchanged.
Unit tests have been added to test this behavior, so the unit test suite should pass.

## Automation
Resolves related issues caught by automation team.

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/TextBox/TextBox.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/TextBox/TextBox.d.ts
@@ -32,6 +32,7 @@ export default class TextBox extends Base {
   content?: string | TextContent[];
   fixed?: boolean;
   marquee?: boolean;
+  displayOnLoad?: boolean;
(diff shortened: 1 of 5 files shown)


erautenberg (commented):
I think this makes sense, but to match other patterns, can we:
- swap the language to me something more like "hideOnLoad" so that it can remain undefined at first and someone has to actively set it to true to make it happen (setting a flag that defaults from true to false always feels odd to me)
- add a control to the TextBox component story for this

ImCoolNowRight (approved):

erautenberg (approved):

PR #365, opened 2023-08-25
fix(Button): remove all unused prefixH and suffixH values

erautenberg:
## Description

In scrubbing the library for other width/height values to alias, I found the prefixH and suffixH values as styles in all extensions of Button. These are completely unused and must have been a mistake in extending Button, or leftover for something we no longer support in Button. Button has getters for prefixW and suffixW that are shorthands to get the width of the component, but that should not be changed.

## Testing

Confirm that Button, ButtonSmall, Control, ControlSmall, and all ListItems seem to work and display the same way as before. Confirm no unit tests or snapshots have changed.

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Button/ButtonSmall.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/Button/ButtonSmall.d.ts
@@ -18,16 +18,10 @@
 
 import lng from '@lightningjs/core';
 import Button, { ButtonStyle } from './Button';
-import { StylePartial } from '../../types/lui';
-
-type ButtonSmallStyle = ButtonStyle;
 
 declare class ButtonSmall<
   TemplateSpec extends Button.TemplateSpec = Button.TemplateSpec,
   TypeConfig extends lng.Component.TypeConfig = lng.Component.TypeConfig
-> extends Button<TemplateSpec, TypeConfig> {
-  get style(): ButtonSmallStyle;
-  set style(v: StylePartial<ButtonSmallStyle>);
-}
+> extends Button<TemplateSpec, TypeConfig> {}
 
-export { ButtonSmall as default, ButtonSmallStyle };
+export { ButtonSmall as default, ButtonStyle };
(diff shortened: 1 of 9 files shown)


soumyaloka (approved):
I actually noticed while updating TS structure for Control component and created a branch for scrubbing these last week and totally forgot to create a PR. Thanks!

packages/@lightningjs/ui-components/src/components/Button/ButtonSmall.d.ts
```diff
@@ -20,10 +20,7 @@ import lng from '@lightningjs/core';
 import Button, { ButtonStyle } from './Button';
 import { StylePartial } from '../../types/lui';
 
-type ButtonSmallStyle = ButtonStyle & {
-  prefixH: number;
-  suffixH: number;
-};
+type ButtonSmallStyle = ButtonStyle;
```
anthony9187: we don't need to re-declare the type here if `ButtonSmall` has the same style type. we can remove this as well as the style accessors since they're the same type as `Button`'s

soumyaloka: Yeah something like this:

```
import Button from './Button';

declare class ButtonSmall<
  TemplateSpec extends Button.TemplateSpec = Button.TemplateSpec,
> extends Button<TemplateSpec, TypeConfig> {}

export default ButtonSmall;
```

erautenberg: @anthony9187 awesome, just updated, let me know if this is what you meant. I still exported the parent Button's style, wasn't sure if that was needed or not.

ImCoolNowRight: Looks like Control needed `ButtonSmallStyle`, so we'll need to either change it to use `ButtonStyle` or add `ButtonSmallStyle` back in if we want a type between the two.

ImCoolNowRight (approved):

PR #368, opened 2023-08-30
fix: update deep merge order for withThemeStyles

joshhowenstine:
## Description

After the withThemeStyles refactor, the merge order of the styles is off. The styles should be applied in the following order:

1. tones
2. modes
3. tone-modes
4. mode-tones

Instead, modes seem to always take precedence over tone-modes. 

## Testing
 Adding these rules to button should be a sufficient test. Should also work at the componentConfig level
  ```
class Button extends lng.Component {
    static _template() {
      return {
        Button: {
          type: ButtonComponent,
          style: {
            mode: {
              unfocused: {
                height: 50
              }
            },
            tone: {
              inverse: {
                mode: {
                  disabled: {
                    height: 500
                  }
                }
              },
              brand: {
                mode: {
                  unfocused: {
                    height: 10
                  }
                }
              }
            }
          }
        }
      };
    }
  }; 

```

### DIFF
--- a/packages/@lightningjs/ui-components/src/mixins/withThemeStyles/utils.js
+++ b/packages/@lightningjs/ui-components/src/mixins/withThemeStyles/utils.js
@@ -405,14 +405,12 @@ export const generateComponentStyleSource = component => {
     for (const toneItem of [
       ...new Set(['neutral', ...Object.keys(tone), ...toneProperties])
     ]) {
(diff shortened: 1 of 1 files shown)


ImCoolNowRight (approved):

soumyaloka (approved):
Tested, works as expected to me!

PR #369, opened 2023-09-05
fix(Icon): do not prevent color prop from being passed to an SVG icon

erautenberg:
## Description

If an SVG is pure white (same as PNGs), a color prop can be passed to it effectively. This removes the old code preventing the color prop from being applied, as well as updates our documentation to no longer specify PNGs. This gives our client teams more flexibility, especially since many would prefer to use SVGs.

## Testing

In the Icon.stories.js file, the SVG story is now using a fill of white, then applying a color the make the circle red on the fly.

## Checklist


joshhowenstine (approved):

packages/@lightningjs/ui-components/src/components/Icon/Icon.js
```diff
@@ -149,10 +149,9 @@ export default class Icon extends Base {
         src: this.icon
       };
     }
-    const supportsColor = !Boolean(svgTag || svgURI);
     const color = getValidColor(this._color || this.style.color);
 
-    if (!supportsColor && color) {
+    if (color) {
       context.warn('icon does not allow colors to be applied to an SVG');
```
anthony9187: should this error message change or be removed? im not sure what case we're catching here

erautenberg: 🤦‍♀️ yes, removing, thank you! Didn't read what the if statement was doing

anthony9187 (approved):

PR #371, opened 2023-09-19
fix(InlineContent): ignore invalid maxLines numbers

THoj13:
## Description
Update the setter for `maxLines` to only accept numbers that are greater than or equal to 1.
Numbers less than 1 will be ignored.
If a number with decimal values is passed in, it will be rounded down to the closest whole number.

## References

LUI-1046
## Testing
On `TextBox / With Inline Content String` story:
1. set `contentWrap: true`
2. set `maxLines` to any positive integer: that number of lines should be displayed
3. set `maxLines` to 0: no truncation should be applied and the full text renders
4. set `maxLines` to a negative number: no truncation should be applied and the full text renders
5. set `maxLines` to decimal number less than 1: no truncation should be applied and the full text renders
6. set `maxLines` to decimal number greater than 1: that number will be rounded down and display that number of lines (ex. 3.5 will render 3 lines)

## Automation
Bug reported by automation. Verify it is resolved with this change.

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
+++ b/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
@@ -444,6 +444,12 @@ export default class InlineContent extends Base {
     return this._customStyleMappings || {};
   }
 
+  _setMaxLines(maxLines) {
+    // only accept positive numbers
+    // round down any decimals to whole numbers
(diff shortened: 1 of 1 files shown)


soumyaloka (approved):
Tested, everything works as expected. Thank you!

anthony9187 (approved):

PR #376, opened 2023-09-26
feat(Tile): Add Icon Support to Tile Component

soumyaloka:
## Description

In this PR, I have added a new method, _updateIcon, to the Tile component. This method adds support for displaying an icon within the Tile and ensures proper alignment based on metadata location, persistent metadata and focused mode.

## References

LUI-1101

## Testing

1. Make sure the icon aligns properly in focused mode when metadatalocation is set to _inset_ icon should show up above metadata and progress bar and when _standard_ should show up above ProgressBar.
2. Icon should be visible in all modes when persistentmetadata is set to true.


packages/@lightningjs/ui-components/src/components/Tile/Tile.js
```diff
@@ -47,6 +48,9 @@ export default class Tile extends Surface {
         },
         Content: {
           mount: 0.5
+        },
+        ProviderLogo: {
+          type: Icon
         }
```
THoj13: Is ProviderLogo being referenced anywhere? It looks like _updateIcon is patching iconSrc to different element, called Icon.

packages/@lightningjs/ui-components/src/components/Tile/Tile.js
```diff
@@ -175,6 +182,26 @@ export default class Tile extends Surface {
...
+    const iconObject = {
+      w: this.style.iconWidth,
+      h: this.style.iconHeight,
+      icon: this.iconSrc,
+      alpha: this._isFocusedMode ? 1 : 0.01,
+      x: this.style.iconX,
+      y:
+        this.metadataLocation === 'inset'
+          ? this.style.iconYInset
+          : this.style.iconYStandard
```
THoj13: Another way I believe you could set the `y` position of the icon would be to calculate its position based on where the MetaData is and subtract by its height to position the icon above it. 
```suggestion
      mountY: this._isInsetMetadata ? 1 : 0,
      y: this._isInsetMetadata
        ? this._metadataY - (this._Metadata ? this._Metadata.h : 0)
        : this.style.iconYStandard
```
By doing it like this, it should also handle if the Tile's height were to increase. The provider logo will always position itself using Metadata's positioning.

packages/@lightningjs/ui-components/src/components/Tile/Tile.js
soumyaloka: Sorry I named it providerLogo first. I will update that.

packages/@lightningjs/ui-components/src/components/Tile/Tile.js
soumyaloka: You are right, Thanks for suggestion. I will update that now.

packages/@lightningjs/ui-components/src/components/Tile/Tile.js
```diff
@@ -175,6 +179,30 @@ export default class Tile extends Surface {
...
+  _updateIcon() {
+    const iconObject = {
+      w: this.style.iconWidth,
+      h: this.style.iconHeight,
+      icon: this.iconSrc,
+      x: this.style.iconX,
+      mountY: this._isInsetMetadata ? 1 : 0,
+      y: this._isInsetMetadata
+        ? this._metadataY - (this._Metadata ? this._Metadata.h : 0)
+        : this.style.iconYStandard,
```
THoj13: If the design is to always display the providerLogo at the bottom of the Tile, I believe `this.style.iconYStandard` can also be replaced with a calculation using the heights and positions of the Tile and ProgressBar. That is part of how the metadata gets its y position when it is inset(`get _metadataY`). It seems like when `metadataLocation` is `standard`, the logo should be positioned at the same place where metadata is positioned when `inset`, so could use that same calculation:
```js
this._progressBarY
    ? this._progressBarY - this.style.paddingYBetweenContent
    : this._h - this.style.paddingY
```

soumyaloka: Good one, yes I can update that, but I would like to confirm with Josh see if this aligns with the required design?
@joshhowenstine

joshhowenstine: Yes, I believe we should have a `paddingYProgress` equivalent for the new icon

THoj13 (approved):

ImCoolNowRight (approved):

joshhowenstine:
image

Looks good! A few adjustments would allow us to be able to adjust just the icon at the theme level. Then perhaps we could add `paddingYIcon: theme.spacer.xs,` to the style file?

soumyaloka:
> image
> 
> Looks good! A few adjustments would allow us to be able to adjust just the icon at the theme level. Then perhaps we could add `paddingYIcon: theme.spacer.xs,` to the style file?

Sorry Josh, just merged this PR, I can create another PR to address your comment.

PR #378, opened 2023-09-26
fix(Tab): ts updates

slhay28:
## Description

Tab and TabBar ts updates

## References

LUI-967


packages/@lightningjs/ui-components/src/components/TabBar/Tab.d.ts
```diff
@@ -49,3 +74,5 @@ export default class Tab extends Surface {
   get _Icon(): lng.Component;
   get _Text(): lng.Component;
 }
+
```
soumyaloka, line 78: Looks good, but can you verify if you have added all the tags because I think content is missing in Tab and I don't see any tags in TabBar.d.ts file?

soumyaloka: minor one, remove type from import statement.

packages/@lightningjs/ui-components/src/components/TabBar/Tab.d.ts
```diff
@@ -49,3 +74,5 @@ export default class Tab extends Surface {
   get _Icon(): lng.Component;
   get _Text(): lng.Component;
```
anthony9187: can we type these as `Icon` and `TextBox`?

anthony9187 (approved):

soumyaloka (approved):

PR #381, opened 2023-10-03
fix(Button): handle prefix arrays and objects with color updates

THoj13:
## Description
The change introduced in this https://github.com/rdkcentral/Lightning-UI-Components/pull/373 resolved an issue of prefix icons/badges flashing from getting re-patched and re-rendered. That change incorrectly assumed that `Button.prefix` would always be an array and relied on looking up an item by index on that array. `Button.prefix` also supports accepting an object. This change allows looking up the desired value for both prefix arrays and prefix objects.

## Testing
n/a - issue that prompted this is only visible in inner source repo

## Automation
n/a

## Checklist


ImCoolNowRight (approved):
Tested by building and copying the `dist` folder into the IS repo. The NavigationLeft story works as expected!

anthony9187 (approved):

PR #385, opened 2023-10-11
fix(FocusManager): call super in _focus and _unfocus without smoothing

ImCoolNowRight:
## Description

This PR breaks out the smooth check in `Base` into a new method, which allows FocusManager to call `super._focus()` and `super._unfocus` to properly update as its mode changes.

## References

N/A

## Testing

- Ensure all FocusManager, NavigationManager, Column, Row, etc. stories all look correct.

## Automation

N/A

## Checklist


joshhowenstine (approved):

packages/@lightningjs/ui-components/src/components/Base/Base.js
```diff
@@ -54,14 +54,20 @@ class Base extends lng.Component {
   _update() {}
 
   _focus() {
-    if (this.shouldSmooth === undefined) this.shouldSmooth = true;
+    super._focus();
```
anthony9187: I'm not sure we need this super call? as far as I can tell `lng.Component._focus()` doesn't exist
https://github.com/rdkcentral/Lightning/blob/master/src/application/Component.d.mts

packages/@lightningjs/ui-components/src/components/Base/Base.js
```diff
@@ -54,14 +54,20 @@ class Base extends lng.Component {
...
+    this._updateShouldSmooth();
     this.queueRequestUpdate();
   }
 
   _unfocus() {
+    super._unfocus();
```
anthony9187: same here as above

packages/@lightningjs/ui-components/src/components/Base/Base.js
ImCoolNowRight: https://github.com/rdkcentral/Lightning/blob/4a3ecc189a7a978790346143e62cb99ac7f1c30c/src/application/Component.d.mts#L916 Is this not it?

anthony9187 (approved):

PR #391, opened 2023-10-19
fix: Replace pointerH and pointerW with pointerHeight and pointerWidth

soumyaloka:
## Description

This PR addresses the problem by replacing the old names inside the Bubble texture, resolving the tooltip size and missing bubble issue

## References

LUI-1156

## Testing

Verify if Tooltip UI looks as expected.


ImCoolNowRight:
Should there be an `aliasProperties` set up for these changes?

Sorry missed your comment, so those texture classes don't use the withThemeStyles mixin so they don't have the alias support.

ImCoolNowRight (approved):

arwehrman (approved):

PR #397, opened 2023-10-31
fix(gitignore): add .gitignore to themePropertyTables directory

ImCoolNowRight:
## Description

Adds a `.gitignore` to the `themePropertyTables` directory to ensure that the directory always exists, but files within it continue to be ignored.

The error, up until now, was that the directory did not exist by the time the runner attempted to write the files.

## References

LUI-849

## Testing

Will have to check to make sure that Actions no longer fail when trying to write `color.md`.

## Automation

N/A

## Checklist

### DIFF
--- a/packages/apps/lightning-ui-docs/scripts/themePropertyTables/.gitignore
+++ b/packages/apps/lightning-ui-docs/scripts/themePropertyTables/.gitignore
@@ -1,0 +1,4 @@
+# Ignore everything in this directory
+*
+# Except this file
+!.gitignore


joshhowenstine:
This should totally work I think. However, as an alternative we could also change line 235 to be synchronous? Honestly, ether approach is fine with me. 

`try {
    fs.mkdirSync(mdFile.substring(0, mdFile.lastIndexOf('/')), { recursive: true });
} catch (err) {
    if (err.code !== 'EEXIST') {
        throw err;
    }
}`

ImCoolNowRight:
> This should totally work I think. However, as an alternative we could also change line 235 to be synchronous? Honestly, ether approach is fine with me.
> 
> `try { fs.mkdirSync(mdFile.substring(0, mdFile.lastIndexOf('/')), { recursive: true }); } catch (err) { if (err.code !== 'EEXIST') { throw err; } }`

I tried this yesterday on the IS repo. It worked fine for me locally, but continued to run into the same error on GitHub. I like keeping this (almost) empty directory around, since we know we'll always need it to exist.

arwehrman (approved):

anthony9187 (approved):

soumyaloka (approved):

PR #403, opened 2023-11-08
fix: calculated height displayed in the inspector for Button and Tile

erautenberg:
## Description

Per, [LUI-1192](https://ccp.sys.comcast.net/browse/LUI-1192) the height of the Button in the inspector is zero. On further inspection of Tile, which also utilizes `get h()`, I figured out that the inspector is not updated unless the `this._h` value is updated and the `_updateDimensions` method is triggered.

For Button:
I had to balance not wanting to manually set "h" and flipping the "hSetByUser" flag and potentially circumventing style/theme updates from changing the Button's dynamic height.

For Tile:
We needed to retain the original size of the Tile _without_ the Metadata via "innerH" and "_h." The final solution was to tap into `_getRenderHeight` and return the value we want, not automatically assume `this._h` is correct.

## Testing

1. Inspect the Button and Tile components in storybook and confirm that the value of height reflects the actual component height.
2. Try manually setting the height of the Button per #399:
>1. In the Button story, trying setting the height via "h" on the component, or nested inside of "style" (both as "h" and "height").
>2. Confirm that the "expanding height" stories in Row and Column now work as expected.

## Checklist


ImCoolNowRight (approved):
As far as I can tell, this is looking good!

arwehrman (approved):

PR #407, opened 2023-11-13
fix(Button): support re patching prefix after undefined

THoj13:
## Description
Previously with the Button component, if the `prefix` were to be defined, then set to undefined, then reset with the original value: the prefix would not be re-rendered. This adds support to toggling a prefix between undefined and a defined value (ex. toggling between displaying and hiding the same icon in the prefix). 

## References
LUI-1178

## Testing
This bug isn't reproducible in storybook as the component is recreated when toggling prefix. A unit test has been added to test the change. Removing the changes in Button.js will result in the unit test failing.

## Automation
Issue found by automation team. Follow up with them after merge/publish.

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Button/Button.js
+++ b/packages/@lightningjs/ui-components/src/components/Button/Button.js
@@ -105,6 +105,7 @@ export default class Button extends Surface {
   }
 
   _updatePrefix() {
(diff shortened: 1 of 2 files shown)


ImCoolNowRight (approved):
I built the package and brought it into IS to test on NavItem. Looks great!

soumyaloka (approved):

PR #416, opened 2023-11-20
fix: allow componentConfig props to properly merge with component props

joshhowenstine:
## Description

Props set in componentConfig are not properly merged and are leaving some style properties when themes are changed at runtime. Also any componentConfig properties should deep merge with properties that are set at the instance level.

## References

[Ticket](https://ccp.sys.comcast.net/browse/LUI-1217)

## Testing

1. Open the Tile component
2. Right click on the iframe window containing the component in storybook and open the webmaster tools
3. Observe that no image scale on artwork is present on focus.
3. `window.CONTEXT.setTheme({componentConfig:{Tile:{artwork:{style:{imageScale:5,},},style:{paddingYProgress:10,paddingYBetweenContent:10,},},},});`
4. After update when focused you should see an imageScale effect when focusing on the Tile
5. Any combination not including the scale property should clear it. `window.CONTEXT.setTheme({componentConfig:{Tile:{artwork:{},style:{paddingYProgress:10,paddingYBetweenContent:10,},},},});`
6. Setting any value on an instance of the component itself should not allow the componentConfig to overwrite it. 

image

## Checklist


ImCoolNowRight (approved):

arwehrman:
I created a build and pulled it into IS. Everything looks ok but there are a few tests failing. Looks like there all variations of Rail so not sure if this related to your changes or not.

erautenberg (changes requested):
I ran a `yarn build` and copied the dist over to IS LUI and the `imageScale` doesn't seem to be applied correctly at all anymore, no matter the theme. In the Tile story in the Xfinity theme, nothing seems to happen when swapping between focus and unfocus when it should be scaling. On the Rail story, if I start on the Xfinity theme, it seems like the first tile in the rail is stuck in the scaled up view, and if I switch to Mosaic, it is still stuck that way as well.

Update: The imageScale in current IS LUI seems totally broken already and is not updating on focus/unfocus. Maybe this was a result of the last update?

PR #417, opened 2023-11-20
fix(NavigationManager): adjust isFullyOnScreen and isComponentOnScreen

erautenberg:
## Description

The `onScreenEffect` in our ColumnContainer and Gallery components was not properly alpha-ing off the rows as they scrolled above and off screen. These adjustments make sure that if a component's _parent_ is animating (specifically inside of a FocusManager), that we use the target transition value to determine if the component is fully on the screen or not, otherwise the `isComponentOnScreen` will return true, even as a row is sliding partially off screen.

## Testing

Will need to build and import into IS LUI to confirm it is working as expected in ColumnContainer and Gallery, otherwise you can make a custom story of Rows nested in Columns and add the following as the Column's `onScreenEffect`
```js
      onScreenEffect: function () {
        if (this.items && this.items.length) {
          this.items.forEach((item, idx) => {
            const alpha =
              item.isFullyOnScreen() || idx >= this.selectedIndex ? 1 : 0;

            if (this.shouldSmooth) {
              item.smooth = { alpha };
            } else {
              item.alpha = alpha;
            }
          });
        }
      }
```

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/docs/Base.mdx
+++ b/packages/@lightningjs/ui-components/src/docs/Base.mdx
@@ -80,7 +80,7 @@ Any component which extends the Base component and uses the `withThemeStyles` mi
 This method accepts a target component, patch object, and optional smooth object. If the component is visible, it will smooth in the smooth object, or fall back to
 the patch object, if not it will apply the patch.
 
-#### isFullyOnScreen({ offsetX: number, offsetY: number }): bool
+#### isFullyOnScreen(\{ offsetX: number, offsetY: number \}): bool
 
 Returns a boolean for whether or not the entirety of the component is rendered within the bounds of the screen size.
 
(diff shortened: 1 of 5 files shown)


packages/@lightningjs/ui-components/src/docs/Base.mdx
```diff
@@ -80,7 +80,7 @@ Any component which extends the Base component and uses the `withThemeStyles` mi
 This method accepts a target component, patch object, and optional smooth object. If the component is visible, it will smooth in the smooth object, or fall back to
 the patch object, if not it will apply the patch.
 
-#### isFullyOnScreen(): bool
+#### isFullyOnScreen({ offsetX: number, offsetY: number }): bool
```
arwehrman: getting a build error on this file
need to add \ in front of both curly braces

arwehrman:
On the Basic story the Button/Tile when focused is slightly hidden at the top. Not related to the changes you made, I think so we could just put a bug ticket in for it.
I created a build of the OS and copied over to IS, tested ColumnContainer, everything works as expected

arwehrman (approved):
Just need to add the \  mentioned in the comments of Base.mdx

ImCoolNowRight (approved):

packages/@lightningjs/ui-components/src/docs/Base.mdx
erautenberg: fixed, thanks!

PR #420, opened 2023-11-21
fix(Artwork): changes one of the parameters to be commented out

arwehrman:
## Description

This pr fixes the issue with `shouldScale` control not working on Artwork. The issue was resolved by commenting out one of the `Artwork.parameters = {....` 
Looking through the commit history I can't seem to track down where we introduced having to parameter assignments but we can't do this, Storybook gets confused on which one to use.  
I'm guessing we had two for a reason at some point. We need to decide which one to keep or we could create two stories.

## References

LUI-1238

## Testing
On OS side, change the `imageScale` prop to any number in the `Artwork.styles.js`, `yarn start`
Got to Artwork controls and toggle `shouldScroll`

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Artwork/Artwork.stories.js
+++ b/packages/@lightningjs/ui-components/src/components/Artwork/Artwork.stories.js
@@ -108,37 +108,6 @@ Artwork.argTypes = {
   }
 };
 
-// Artwork.parameters = {
-//   argActions: {
-//     foregroundSrc: (foregroundSrc, component) => {
-//       component.tag('Artwork').foregroundSrc =
-//         'none' !== foregroundSrc ? foregroundSrc : undefined;
-//     },
-//     srcCallback: (active, component) => {
-//       if (active) {
-//         // Accepts a regular function or function that returns a promise
-//         component.tag('Artwork').patch({
-//           src: '8501866671289235112',
-//           srcCallback: ({ w, closestAspectRatio, src }) => {
-//             return new Promise(resolve => {
-//               setTimeout(() => {
-//                 resolve(
-//                   `https://myriad.merlin.comcast.com/select/image?entityId=${src}&width=${w}&ratio=${closestAspectRatio}&rule=noTitle`
-//                 );
-//               }, 500);
-//             });
-//           }
-//         });
-//       } else {
-//         component.tag('Artwork').patch({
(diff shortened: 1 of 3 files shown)


arwehrman:
Sanity tests fail if we comment out the second one, which makes sense after further investigation.
 It appears that when both Artwork.parameters are available, it uses the second one and that's what the Sanity test are built off of

packages/@lightningjs/ui-components/src/components/Artwork/Artwork.stories.js
```diff
@@ -108,36 +108,36 @@ Artwork.argTypes = {
...
+//         });
+//       } else {
+//         component.tag('Artwork').patch({
+//           src: 'https://myriad.merlin.comcast.com/select/image?entityId=8501866671289235112&width=400&ratio=3x4&rule=noTitle',
+//           srcCallback: undefined
+//         });
+//       }
+//     }
+//   }
+// };
```
ImCoolNowRight: Can we remove this instead? It looks like this was what it was originally for IS (since it's pointing to Myriad) and was a reference for what is below (which seems to be the same thing, just different `src`s.

arwehrman: Yep. I just didn't want to delete it until someone confirmed that we didn't need both for some reason.

erautenberg (approved):

erautenberg:
Since the imageScale is undefined, it's not showing up in the theme styles panel, so we don't actually see shouldScale doing anything in the story. Can we set this to 1 by default?

ImCoolNowRight (approved):

packages/@lightningjs/ui-components/src/components/Artwork/Artwork.styles.js
```diff
@@ -29,7 +29,7 @@ export const base = theme => ({
   fallbackSrc: undefined,
   fillColor: theme.color.overlay,
   gradientColor: theme.color.material,
-  imageScale: undefined,
```
ImCoolNowRight, line 32: Tile has a failing test that needs to be updated because of this change.

arwehrman: Test is passing now. Updated the test to check for 1 since that is the Artwork's default

PR #421, opened 2023-11-27
feat(Keyboard): calculate key size based on base key width

THoj13:
## Description
This changes how the width of `Key` components is calculated so that each Key's width is a multiple of the smallest key width (plus the spacing between keys). This results in the left and right borders of each key aligning with the borders of keys above and below it.

## References
LUI-1191

## Testing
1. View the  Keyboard stories
2. All keys widths should be multiples of the smallest key's width. A way to see this is to view the keyboard not center aligned and the left/right borders should align. 

## Automation
If automation is checking widths of keys, there may be changes required for this.

## Checklist


erautenberg (changes requested):
Overall looks good! I'd just like to rearrange where things live and ensure we don't create an breaking changes in the styles. So `sizes` should still live there. Instead, it will just include the multiple number, like this:
```js
sizes: {
  sm: 1,
  md: 2,
  lg: 3,
  xl: 4,
  xxl: 5
}
```
Then in the Key class, you would do `this.style.baseWidth * this.style.sizes[this.size] + this.style.keySpacing * (this.style.sizes[this.size] - 1)`

erautenberg:
Also, can the user pass in a hard-set width? Like if size is a number (check out columnSize in Gallery). Or, could they just pass in a width directly to the Key and apply that? Maybe like the fixedWidth part of Button?

THoj13:
> Also, can the user pass in a hard-set width? Like if size is a number (check out columnSize in Gallery). Or, could they just pass in a width directly to the Key and apply that? Maybe like the fixedWidth part of Button?

Good point. Key extends Button, so I think it'd makes sense for it to follow the same API (setting `fixed: true` to allow defining `w` on it).

packages/@lightningjs/ui-components/src/components/Key/Key.mdx
```diff
@@ -52,23 +52,24 @@ class Basic extends lng.Component {
...
 
 ### Style Properties
 
-| name       | type             | description                                                                            |
-| ---------- | ---------------- | -------------------------------------------------------------------------------------- |
-| iconWidth  | number           | width set to all icons                                                                 |
-| iconHeight | number           | height set to all icons                                                                |
-| minWidth   | number           | used as a fallback width if no width is passed into the component and `fixed` is false |
-| paddingX   | number           | space between the button horizontal edge and the content                               |
-| sizes      | object           | used to determine the width of key, contains the properties `sm`, `md`, `lg`, `xl`     |
```
erautenberg, line 70: I think the `sizes` line needs to be brought back

packages/@lightningjs/ui-components/src/components/Key/Key.mdx
```diff
@@ -52,23 +52,24 @@ class Basic extends lng.Component {
...
+| ---------- | -------- | -------- | --------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| icon       | string   | false    | null      | url for icon                                                                                                                                                                                                          |
+| onEnter    | function | false    | undefined | if defined, will be invoked when enter is pressed while focused on the `Key` component. When defined, the `$onSoftKey` and `$toggleKeyboard` events will not be emitted from the key (see Events section below)       |
+| size       | string   | false    | sm        | Used to determine the with of the key. Accepts: `sm`, `md`, `lg`, `xl`, `xxl`.                                                                                                                                        |
```
erautenberg: this can be removed/move back to style props

THoj13:
> Also, can the user pass in a hard-set width? Like if size is a number (check out columnSize in Gallery). Or, could they just pass in a width directly to the Key and apply that? Maybe like the fixedWidth part of Button?

Good call out. I ended up following the same API for a fixed width Button (9623561b8e1fd35dac7f0aabe727bd626e24aa49)

ImCoolNowRight (commented):
Everything looks great! Just realized that the TS definitions are going to need to be updated for the new properties and styles.

erautenberg (approved):

ImCoolNowRight (approved):

PR #425, opened 2023-12-01
fix(TextBox): if content changes from array to string (InlineContent to text), reset marquee props

erautenberg:
## Description

If the TextBox's content changes from InlineContent to a string, the Marquee was not properly updated.

## Testing

In the TextBox's InlineContent story, add `marquee: true` and something like this to ensure you swap from an array to plain text:
```js
    _init() {
      setTimeout(() => {
        this.tag('TextBox').content = 'plain text';
      }, 2000);
    }
```
Previously, the text would just disappear, now it should update.

## Checklist


ImCoolNowRight (approved):
Works on the Gallery PR.

THoj13 (approved):

PR #428, opened 2023-12-07
fix(withThemeStyles): w/h setters don't need to clear style cache

ChrisArasin:
## Description

- Currently, in `withThemeStyles`, the w/h setters clear the StyleManger style cache. However, the style calculation doesn't change based on the w/h props. 
- Instead, we can just run the component update cycle without clearing the style cache. 

## Testing

- In Tile story, change the itemLayout props to update the item w/h, Tile should still be styled properly. 
- If anyone can think of anything else this may impact, let me know. 
- Slider and TabBar both needed minor fixes. Ensure they render properly on init and function as expected.

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/ScrollWrapper/ScrollWrapper.js
+++ b/packages/@lightningjs/ui-components/src/components/ScrollWrapper/ScrollWrapper.js
@@ -40,7 +40,7 @@ export default class ScrollWrapper extends Base {
         type: ScrollSlider,
         vertical: true,
         signals: {
-          onChange: '_updateScrollContainerSize'
+          onSizeChange: '_updateScrollContainerSize'
         },
         announce: ' '
       }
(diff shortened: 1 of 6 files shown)


anthony9187 (approved):
spot tested a verity of components and it all seems good to me

ImCoolNowRight:
> - In Tile story, change the itemLayout props to update the item w/h, Tile should still be styled properly.

Can you clarify this? Changing the `itemLayout` in Tile.stories.js keeps the Tile formatted the same as before the change. Is that what you mean by "Tile should still be styled properly"?

ChrisArasin:
> > * In Tile story, change the itemLayout props to update the item w/h, Tile should still be styled properly.
> 
> Can you clarify this? Changing the `itemLayout` in Tile.stories.js keeps the Tile formatted the same as before the change. Is that what you mean by "Tile should still be styled properly"?

Ah yeah, that just triggers a w/h change on the tile, So testing that the tile displays properly after that change happens.

ImCoolNowRight (approved):

ChrisArasin:
I believe this is introducing an issue where a couple tests sometimes fail and sometimes pass. Feels like an issue with awaiting test spies.

Edit: Should be resolved.

PR #430, opened 2023-12-11
fix(withExtensions): Performance refactor

ChrisArasin:
## Description

 withExtensions currently runs on `_construct` for every instance of a component class. This updates the extension creation process to only modify component prototypes, so the createExtensions workflow only needs to run once for all instances of a component class.

- Adds additional tests
- Adds a note to the readme about the `_construct` hook not being supported in extensions. This is not a new limitation, it was just previously not documented. 
- Adds timestamp to theme when processed for tracking purposes
- **Also question**: there are technically some type changes, but only on underscored methods. Let me know how we semver that?
- Note on **subtheme**: I don't believe the current versions handles running the cleanup methods on subtheme events, so I haven't gotten into that. May be an outstanding feature?

## Testing

Tricky to test this in open source repo... Need to apply some extensions and make sure they are functioning properly. 

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/docs/ThemingExtensions.mdx
+++ b/packages/@lightningjs/ui-components/src/docs/ThemingExtensions.mdx
@@ -145,6 +145,11 @@ If we find through A/B testing, the results are not in favor of the vibrant colo
 
 ## Things to Keep in Mind
 
+### `_construct()` usage restriction
+
+Because extensions are applied at runtime on `_contruct` of the component instance, they cannot properly support a `_construct` method override in the extension so must be avoided.
+Other lifecycle hooks are available and supported, such as `_init` and `_attach`.
+
 ### `super._update()`
 
 If we consider extensions as the child of the existing component it builds off, the `super` keyword allows us to access the properties and methods in the existing/parent component. The `_update` method is an extremely important lifecycle event since it ensures everything is up to date. Together, `super._update()` grants us access to the entire `_update` lifecycle event of the parent component.
--- a/packages/@lightningjs/ui-components/src/mixins/withExtensions/index.d.ts
+++ b/packages/@lightningjs/ui-components/src/mixins/withExtensions/index.d.ts
@@ -34,6 +34,8 @@ export interface WithExtensions {
   _instanceNeedsReset(): boolean;
   _resetPrototype(): void;
   _handleThemeChange(): void;
+  _markInstanceUpToDate(): void;
+  _markClassUpToDate(): void;
 }
 
 export interface WithExtensionsConstructor {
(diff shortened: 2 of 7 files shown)


packages/@lightningjs/ui-components/src/mixins/withExtensions/index.d.ts
```diff
@@ -19,24 +19,28 @@
...
-  get _extensionsApplied(): boolean;
 
   _resetComponent(): void;
   _calculateComponentExtensionLength(): number;
   _createExtension(): void;
-  _createExtensionClass(): unknown;
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  _createExtensionAliases(obj: Record<string, any>): object;
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  _setComponentAliases(aliasObj: Record<string, any>): void;
```
ImCoolNowRight, line 36: We've been adding `@deprecated` to a comment block above things we've removed to warn users that they're no longer supported: https://github.com/rdkcentral/Lightning-UI-Components/blob/6f6c1dfd6d0a056fe41f0aa95ca63a6ab2040861/packages/%40lightningjs/ui-components/src/components/Tile/Tile.d.ts#L34

ImCoolNowRight (dismissed):
Tested in IS Storybook on components that have FocusRing and DropShadow extensions. Everything's looking good!

packages/@lightningjs/ui-components/src/docs/ThemingExtensions.mdx
```diff
@@ -145,6 +145,11 @@ If we find through A/B testing, the results are not in favor of the vibrant colo
 
 ## Things to Keep in Mind
 
+### `_construct()` usage restriction
+
+Because extensions are applied at runtime on `_contruct` of the component instance, they cannot properly support a `_construct` method override in the extension so must be avoided.
+Other lifecycle hooks are available and supported, such as `_init` and `_attach`.
```
ChrisArasin, line 151: An additional thought here. For components that are created before a theme is applied, functionality added in _init and _attach (lifecycle hooks before _update) would not be run. Any event listener setup, etc. could not be guaranteed to happen. 

Our internal extensions essentially only rely on _update. I wonder if this is almost formalized into a requirement for the extensions classes?

Another option could be that a theme change that has a different set of extensions somehow requires re-initializing any current view of the app. That way we can expect everything to be torn down, and _init/_attach hooks could run with a new extensions applied. 

I think we probably prefer the former at the moment, as live-swapping themes that modify the look of the app seems like the general vision for the feature, even if in practice most will just set a theme once at init time. It could be a "here's how it works, do with it what you will" type of documentation as well...

joshhowenstine:
I discovered a unique scenario in which it appears that when you expand a component beyond the boundaries of the UI component library, it may not consistently recognize the extension. For instance, in my application, I have a "TestTile" that extends the "Tile" component. When I apply a "FocusRing" extension to the "Surface," it sometimes fails to load as expected on "TestTile"

ctoddy:
Closing, no recent updates.

PR #433, opened 2023-12-14
fix(Metadata): only create text elements when needed

ChrisArasin:
## Description

MetadataBase instantiates Textbox components for Title, Description, and Details, and an Icon component for the Logo regardless of whether or not these properties are set. Tiles often don't need all of these elements. This change waits to create them until they're needed. 

## Testing

- Ensure MetadataBase and MetadataTile render properly. 
- Set up a Row or Column of Tiles with with Text/Description properties that will require Marquee. Ensure marquee behavior is correct as you navigate across the row. 

## Checklist


packages/@lightningjs/ui-components/src/components/Tile/Tile.js
```diff
@@ -568,14 +568,7 @@ export default class Tile extends Surface {
   }
 
   _getMetadataLocation() {
```
ChrisArasin, line 577: Faster not to modify style here. Not sure there would be any case it's preferable to do it this way though?

arwehrman (approved):

arwehrman:
Did a `yarn build`, copied build files to IS node_modules for @lightningjs/ui-components test out changes on inner source.
Looks good. Emily did do some updates to Tile that don't look like they are merged in this PR, so might want to stress test components after rebase to make sure everything works as expected.

packages/@lightningjs/ui-components/src/components/MetadataBase/MetadataBase.js
```diff
@@ -33,41 +33,23 @@ class MetadataBase extends Base {
...
       }
+      // Logo: {
+      //   flexItem: false,
+      //   type: Icon
+      // }
     };
   }
 
-  _init() {
-    this.requestUpdate(true);
```
ChrisArasin, line 68: This did not seem necessary, but if anyone knows the origins?

packages/@lightningjs/ui-components/src/components/Tile/Tile.js
anthony9187: it may have been to keep styles in sync with props but im not sure if that's correct/relevant anymore?

anthony9187 (approved):

packages/@lightningjs/ui-components/src/components/Tile/Tile.js
anthony9187: @erautenberg:
> So this makes sure that if the user sets it on the component and not nested in styles, it updates the styles. And we have to do this.style = {} instead of this.style.prop = '' because there are no individual setters for properties nested in style that would trigger a new update lifecycle, so you have to set on the style itself

resolving this per our conversation

packages/@lightningjs/ui-components/src/components/MetadataBase/MetadataBase.js
erautenberg: 🤔 seems like very old leftovers to me. @ImCoolNowRight any thoughts? If everything seems to render as expected without it in Tile and Cards, it's probably OK

ImCoolNowRight: My best guess is that this was from some weird formatting bug that could only be solved with an extra update cycle. It's looking fine now, so it probably got resolved and this just got left in because I didn't make a note of when to remove it.

packages/@lightningjs/ui-components/src/components/MetadataTile/MetadataTile.test.js
```diff
@@ -67,11 +67,15 @@ describe('MetadataTile', () => {
 
   it('hides details if description is provided', async () => {
     metadataTile.details = 'details';
-    await metadataTile.__updateSpyPromise;
+    await metadataTile.__detailsLoadedpyPromise;
```
arwehrman: I think there is an S missing in the `_detailsLoadedpyPromise`

ChrisArasin: womp. Thanks!

PR #437, opened 2023-12-19
fix(Tile): remove extra patch call before calling animate to avoid stutter effect

erautenberg:
## Description

Calling the patch, then animating was causing the text to jump when inset. This video shows the before and after.

https://github.com/rdkcentral/Lightning-UI-Components/assets/15019089/d36abacc-a5b8-4d7a-b67a-da1c70f6cd47

## Testing

You would have to test with inset metadata, potentially only in IS LUI on the new Metadata component.

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Tile/Tile.js
+++ b/packages/@lightningjs/ui-components/src/components/Tile/Tile.js
@@ -598,9 +598,6 @@ export default class Tile extends Surface {
 
       return;
     }
-    // if none of the above apply patch in metadataPatch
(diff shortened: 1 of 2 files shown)


ImCoolNowRight (approved):

arwehrman:
Looks good. I did notice that Description and Title don't update when you change them in the controls, but I'm pretty sure this is Storybook issue and not related to the changes you made. 
I wouldn't hold up this PR for that

arwehrman (approved):

PR #461, opened 2024-01-30
feat(Extensions): prototype extensions refactor

ChrisArasin:
## Description

Prototype implementation of applying extensions at setTheme time to Component prototypes rather than at instance construct time.

Extensions take an array of imported component references rather than component names
```
const extensions = [
  {
    components: [Surface],
    extension: WithRedRect,
  },
]
```

Extensions are just classes, and super calls are replaces by this custom next implementation 
```
class WithRedRect {
  _update(next) {
    // super calls are replaced with next
   // this can be bound during applyExtensions instead, but think this may be technically faster. 
     if (next) {
      next.apply(this)
    }
    this.patch({
      PopupRed: {
        rect: true,
        w: 100,
        h: 100,
        x: 10,
        y: 10,
        color: 0xffff0000,
        zIndex: 100,
      },
    })
  }

  _cleanupExtension() {
    this.patch({
      PopupRed: undefined,
    })
  }
}
```

Results in roughly 15-20% faster component construct/init time. 
Perf. was measured creating a column of 4 rows of 8 tiles. 

Before: 

image

After: 

image
## References
https://ccp.sys.comcast.net/projects/LUI/issues/LUI-1329

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/globals/context/applyExtension.js
+++ b/packages/@lightningjs/ui-components/src/globals/context/applyExtension.js
@@ -1,0 +1,75 @@
+const SUFFIX = '__ORIGINAL__';
+
+// If applyExtensions happens after component has been intialized, need to trigger component update cycle after theme is applied
+// This would happen with the themeUpdate event
+export function applyExtension(targetClass, extensionClass) {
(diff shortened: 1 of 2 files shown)


ctoddy:
Closing, no recent updates.

PR #465, opened 2024-02-15
fix(Storybook): set theme and props before render

ChrisArasin:
## Description

During some keyboard layout debugging, I encountered a couple glitches that could be attributed to storybook rather than the component.

**With the following logs set in Keyboard**: 

storybook-log

**It was observed that:**

1. An initial render was happening without the desired props
2. a setTheme event was triggering after the component's first update. 

storybook-before

After these changes, no setTheme event fires after the component mounts, and the props are set as expected on initial render. 

storybook-after

## Testing

This is not a Component change, should just confirm storybook is working as expected. 

## Checklist

### DIFF
--- a/packages/apps/lightning-ui-docs/.storybook/addons/decorators/withLightning.js
+++ b/packages/apps/lightning-ui-docs/.storybook/addons/decorators/withLightning.js
@@ -130,11 +130,24 @@ export const withLightning = (
             this._refocus(); // Force Lightning to reset focus
           }
           _setup() {
-            // This ensures the component as it's args before the first update cycle. Fixes some flashing in storybook.
-            // TODO: confirm if this can be the whole args object or if we have to process it somehow
-            this.componentTarget.patch({
-              ...args
-            });
+            // This ensures the component has its args before the first update cycle.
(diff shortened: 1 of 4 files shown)


packages/apps/lightning-ui-docs/.storybook/addons/decorators/withLightning.js
```diff
@@ -130,6 +130,11 @@ export const withLightning = (
             this._refocus(); // Force Lightning to reset focus
           }
           _setup() {
+            // This ensures the component as it's args before the first update cycle. Fixes some flashing in storybook.
+            // TODO: confirm if this can be the whole args object or if we have to process it somehow
+            this.componentTarget.patch({
+              ...args
```
ChrisArasin: Is there anything we have to do to args here? It appears to just contain the props coming from storybook, but I see there is some conditional handling of it below...

ChrisArasin:
Only thing I didn't look at was argActions... Not entirely sure what they are? Could look at bringing that up into the earlier property patch as well.

erautenberg (approved):

ImCoolNowRight (approved):

PR #467, opened 2024-02-16
fix: updates to optimize tree shaking

joshhowenstine:
## Description

This PR eliminates all re-exports from the codebase and replaces them with direct file references. The primary goal of this change is to improve our project's compatibility with tree shaking. By directly referring to files, we can ensure that only the code actually used in the project is included in the final bundle, which will lead to smaller bundle sizes and better performance in production environments.

## References

NO-JIRA

## Testing

Testing if tree shaking is working properly takes a few steps. You will need to utilize an existing or new project to see the results.

1. Ensure Correct Webpack Configuration: Check that your Webpack configuration is optimized for tree shaking. This includes:

- Utilizing ES6 module syntax (import and export) throughout your project.
- Setting mode to 'production' in your Webpack configuration to enable optimizations like tree shaking.
- Confirming that the sideEffects property in your package.json is properly configured to false or accurately lists fles/modules that do have side effects, aiding Webpack in excluding unused modules.

2. Build the Project: Execute the Webpack build process using npm run build or your project's equivalent script to generate the production bundle. 

3. Review Output: After the build, review the output to ensure there are no errors or warnings related to the tree shaking process or the direct file references. This step is crucial to verify that the modifications have not introduced any build issues. The final bundle should only include Lightning UI components that have been used in your project. All others should no longer be present in the final code.

## Automation

All automation tests should pass as expected. This should have no effect on exported components or functions.

## Checklist


ImCoolNowRight:
I've built this, brought it into inner source's `develop` branch, built that, and then imported into AirUI. When I built the app, I'm still seeing components that certainly aren't being used showing up in the final bundle. We can pair on this when you're back in the office, but I think either I missed something in testing, or something in the PR needs adjusting.

packages/@lightningjs/ui-components/package.json
```diff
@@ -27,12 +27,14 @@
     "test:watch": "NODE_OPTIONS=--experimental-vm-modules jest --no-coverage --watch --colors"
   },
   "peerDependencies": {
-    "@lightningjs/core": "2.11.0"
+    "@lightningjs/core": "^2.7.0"
```
ImCoolNowRight: What does this need to be rolled back for? It seems like it's still resolving to the same version as before, and I think we needed to keep this locked at 2.11.0 until we looked more into the issues related to #400.

packages/@lightningjs/ui-components/src/globals/context/index.d.ts
```diff
@@ -0,0 +1,33 @@
+type ThemeObject = any;
```
ImCoolNowRight: Is this ok to be `any`, or should it at least be an `object`?

packages/@lightningjs/ui-components/src/mixins/withThemeStyles/utils.js
```diff
@@ -354,10 +354,10 @@ export const generateSolution = (
   const solution = {};
 
   const uniqueModes = getUniqueProperties([
+    'unfocused', // Unfocused must be at the end for proper fallback since base === 'unfocused' in many cases
```
ImCoolNowRight: Is this comment no longer true?

packages/@lightningjs/ui-components/src/textures/index-remove.js
ImCoolNowRight, line 1: Is this file supposed to be removed now?

ImCoolNowRight (commented):
Super minor points of clarification, but functionally, I think everything's good to go with this!

packages/@lightningjs/ui-components/package.json
joshhowenstine: Will update, thanks!

joshhowenstine: What if we did something like this?

`"@lightningjs/core": ">=2.0.0 <=2.11.0"`

https://docs.npmjs.com/cli/v10/commands/npm-install

That way the consuming application does not have to be locked to a specific version but does not hit the version that has breaking changes.

packages/@lightningjs/ui-components/src/globals/context/index.d.ts
joshhowenstine: Definitely should be object. Thanks

packages/@lightningjs/ui-components/src/mixins/withThemeStyles/utils.js
joshhowenstine: Accidental, should not have been committed. Reverted.

packages/@lightningjs/ui-components/src/textures/index-remove.js
joshhowenstine: Had multiple files renamed like this that I failed to remove. They should all be cleaned up now.

ImCoolNowRight (approved):
Everything looks good!

slhay28 (approved):

PR #468, opened 2024-02-16
fix(NavigationManager): fix render order and storybook glitches

ChrisArasin:
## Description

- Allows NavigationManager children components to run layout/updates before the parent does. This prevents a mis-rendered stack of children on the first render of column/row.
- Also includes Storybook render glitch fixes to normalize storybook rendering. These changes are described here: https://github.com/rdkcentral/Lightning-UI-Components/pull/465
- It helps to evaluate this PR, but we can merge separately or all together here. 
- Fixes the use of the centerKeys prop in keyboard to fix a mis-rendered first draw. 

## References

Related to: [LUI-1328](https://ccp.sys.comcast.net/browse/LUI-1328)

## Testing

- Ensure Column/Row based layouts are functioning properly. Items should draw in place and not stacked on top of each other when chrome is throttled down.

## Checklist


packages/@lightningjs/ui-components/src/components/Row/Row.test.js
```diff
@@ -431,31 +431,31 @@ describe('Row', () => {
       done();
     });
 
-    it('should pass on screen items to onScreenEffect', done => {
+    it('should pass on screen items to onScreenEffect', async () => {
       row.w = 200;
+      row.clipping = true;
```
ChrisArasin, line 436: This test needs clipping to pass as expected. I'm not sure why it was passing before...

packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.js
```diff
@@ -114,6 +113,9 @@ export default class FocusManager extends Base {
       x: this.itemPosX,
       y: this.itemPosY
     });
+    if (this._lazyItems) {
+      this._lazyItems = null;
```
erautenberg, line 120: does it matter that this is null instead of an empty array

ChrisArasin: Yeah, this is actually broken on develop now too. NavigationManager's `appendItems()` method does this check, so if that array exists when set items is called they just get pushed there and not rendered.

```
  appendItems(items = []) {
    this.shouldSmooth = false;

    if (this._lazyItems) {
      this._lazyItems.push(...items);
      return;
    }
 ```

File: packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.d.ts [hunk 3]
```diff
@@ -176,8 +176,9 @@ declare class FocusManager<
   /**
    * Selects next item. If this.wrapSelected=true, will select the first element in the list if focus is currently on the last item.
    * emits the `selectedChange` signal
+   * @param forceSmoothValue allows setting the shouldSmooth value before the select update
    */
-  selectNext(): void;
+  selectNext(forceSmoothValue?: boolean): void;
```
erautenberg: mind updating FocusManager.mdx too? Line 162 has selectNext

ChrisArasin: yep, gave it a better name too

ImCoolNowRight:
All the stories I tested are looking good, but I did see that the TextPanel portion of GridOverlay still has a Column that is showing the TextBoxes stacked when first rendering:


https://github.com/rdkcentral/Lightning-UI-Components/assets/2678766/f5974b6d-b632-461a-9c15-af839961ef04

ChrisArasin:
> All the stories I tested are looking good, but I did see that the TextPanel portion of GridOverlay still has a Column that is showing the TextBoxes stacked when first rendering:

So this would be an example of where waitForDimensions may still be relevant. In this case, the items don't have a height until the text textures are rendered and the textboxes update. If they had a height set beforehand, then would not stack

ImCoolNowRight (approved):

packages/@lightningjs/ui-components/src/components/FocusManager/FocusManager.js
erautenberg: *will handle error checking for this in a separate ticket

erautenberg (approved):

PR #477, opened 2024-02-22
fix(NavigationManager): if an itemPosX or itemPosY are set, apply offset to scrolling

erautenberg:
## Description

The `itemPosX` and `itemPosY` properties allow the user to offset the `Items` container inside of NavigationManagers. However, these values were never applied to the scrolling logic, so even if the container had been offset, once the user started scrolling, this information would be wiped out entirely.

## Testing

Columns:
- Add `itemPosY: 300` (or any value) to the Column stories
- Observe that the Column has shifted down as expected, and when scrolling, the shift is still in effect
- Confirm the above with
  - regular scrolling (no flags set)
  - alwaysScroll set to true

Rows:
- Add `itemPosX: 300` (or any value) to the Row stories
- Observe that the Row has shifted right as expected, and when scrolling, the shift is still in effect
- Confirm the above with
  - regular scrolling (no flags set)
  - alwaysScroll set to true
  - lazyScroll set to true

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Column/Column.js
+++ b/packages/@lightningjs/ui-components/src/components/Column/Column.js
@@ -82,20 +82,14 @@ export default class Column extends NavigationManager {
         .y;
       const smoothObj = [
         -(scrollItem || this.Items.childList.first).transition('y')
-          .targetValue + (scrollItem === this.selected ? scrollOffset : 0),
+          .targetValue +
+          (scrollItem === this.selected ? scrollOffset : 0) +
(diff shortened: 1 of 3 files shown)


ImCoolNowRight (approved):
Tested Column, Row, TitleRow, and ControlRow. All's looking good!

ImCoolNowRight (approved):
Re-approving just because I think this is the stable point for it.

ChrisArasin (approved):

PR #481, opened 2024-03-01
fix: remove circular dependency warnings when generating bundled ts file

joshhowenstine:
Fixes circular dependency warnings when running `yarn build`


ImCoolNowRight (approved):
Just need to run the linter, but otherwise, it looks good!

slhay28 (approved):

PR #488, opened 2024-03-16
feat: adding ‘retainSelection’ to TabBar

archaica:
## Description

New 'retainSelection' property allows selected tab to remain ‘selected’ in mode even when the TabBar loses focus. That way, if the content is not using the 'TabContent' area in TabBar, we can maintain the same sort of relationship between 'selected' content and retain the visual identity of the selected tab.

## Testing

I added a test to verify this change.

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/TabBar/TabBar.test.js
+++ b/packages/@lightningjs/ui-components/src/components/TabBar/TabBar.test.js
@@ -354,4 +354,16 @@ describe('TabBar', () => {
       expect(tabBar._Tabs.selectedIndex).toBe(1);
     });
   });
+
+  it('should optionally allow the selected tab to remain selected when tab bar loses focus', () => {
+    expect(tabBar._Tabs.items[0].mode).toBe('focused');
+    expect(tabBar._Tabs.items[1].mode).toBe('unfocused');
+
+    tabBar.retainSelection = true;
+    tabBar.mode = 'unfocused';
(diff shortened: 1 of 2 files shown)


joshhowenstine (approved):
Looks great, thanks for contributing!

packages/@lightningjs/ui-components/src/components/TabBar/TabBar.test.js
```diff
@@ -354,4 +354,16 @@ describe('TabBar', () => {
...
     });
   });
+
+  it('should optionally allow the selected tab to remain selected when tab bar loses focus', () => {
+    expect(tabBar._Tabs.items[0].mode).toBe('focused');
+    expect(tabBar._Tabs.items[1].mode).toBe('unfocused');
+
+    tabBar.retainSelection = true;
+    tabBar.mode = 'unfocused';
+    tabBar._updateTabs();
```
anthony9187: in general, we try to avoid directly calling `_update<Name>` methods directly as this doesn't simulate the component update cycle - instead we use `testRenderer.forceAllUpdates()`. You can see it used in the tests here for example https://github.com/rdkcentral/Lightning-UI-Components/blob/develop/packages/%40lightningjs/ui-components/src/components/Button/Button.test.js

ImCoolNowRight (approved):

archaica:
'Text execution failed' is odd; it succeeded locally

PR #490, opened 2024-03-19
feat: add Subtitle Support to MetadataBase Component

soumyaloka:
## Description

In this PR, I have added support for a new subtitle property in the MetadataBase component to accommodate flex design requirements.

## Testing

Navigate to MetadataBase component and verify that it includes a subtitle below the title.


packages/@lightningjs/ui-components/src/components/MetadataBase/MetadataBase.d.ts
```diff
@@ -67,11 +67,6 @@ declare namespace MetadataBase {
      * sets the marquee for Title and Description to the same value so they sync
      */
     marquee?: boolean;
-    /**
-     * @deprecated
-     * relevant content data in the middle
-     */
-    subtitle?: string;
```
ImCoolNowRight, line 74: Was there a reason to not add the subtitle back into MetadataBase and only into MetadataCardContent?

If we're keeping it the way you have it now, we might want @anthony9187's thoughts on how to properly label `subtitle` being deprecated on MetadataBase but being added to MetadataCardContent.

soumyaloka: Hey, Nothing specific but MetadataCardContent is more appropriate for flex use case.

ImCoolNowRight: Can we move this and have `subtitle` added back to MetadataBase, implement it in MetadataCardContent, and leave it available but unused in the rest of the Metadata components?

anthony9187: I feel like it would be confusing to have it available but unused in other Metadata variants

ImCoolNowRight: Sorry, when I said "unused", I meant that we would just not update the stories to use the new property since we don't have designs for it. But if users want it, the components would have the property and would have it implemented properly.

I'm fine with leaving it as is since this is the only case (so far) of the subtitle being needed.

ImCoolNowRight (commented):
Everything looks good, but can we just get a test added to cover the Subtitle?

packages/@lightningjs/ui-components/src/components/MetadataBase/MetadataBase.d.ts
anthony9187: ahhh gotcha. yeah if the code supports it then im all for it

ImCoolNowRight (dismissed):

packages/@lightningjs/ui-components/src/components/MetadataCardContent/MetadataCardContent.js
```diff
@@ -38,6 +38,9 @@ export default class MetadataCardContent extends MetadataBase {
         Title: {
           type: TextBox
         },
+        // Subtitle: {
```
soumyaloka: I will remove this commented code

ImCoolNowRight (approved):

packages/@lightningjs/ui-components/src/components/MetadataBase/MetadataBase.js
```diff
@@ -186,6 +189,38 @@ class MetadataBase extends Base {
...
+        },
+        1
+      );
+    }
+
+    this._Subtitle.patch({
+      content: this.subtitle,
+      marquee: this.marquee,
+      style: {
+        textStyle: {
```
slhay28, line 214: my only thoughts is that the subtitle overlaps with the icons in the details. Should we make the lineheight a little bit larger so this doesn't overlaP?

ImCoolNowRight: I think we might want to leave this as is. The InlineContent uses the line height of the text, so I think the Icon being taller is a side effect of how InlineContent is set up and not necessarily of how it is being used here.

slhay28 (approved):

PR #494, opened 2024-03-26
fix: update definition file for context

joshhowenstine:
## Description

Fixes ThemeObject type for context

### DIFF
--- a/packages/@lightningjs/ui-components/src/globals/context/index.d.ts
+++ b/packages/@lightningjs/ui-components/src/globals/context/index.d.ts
@@ -1,12 +1,12 @@
-type ThemeObject = object;
+type ThemeObject = any;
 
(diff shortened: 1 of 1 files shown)


packages/@lightningjs/ui-components/src/globals/context/index.d.ts
```diff
@@ -1,12 +1,12 @@
-type ThemeObject = object;
+type ThemeObject = any;
```
ImCoolNowRight, line 2: Probably makes sense to have `/* eslint-disable @typescript-eslint/no-explicit-any */` at the top for these lint errors.

joshhowenstine: Good call. Added

ImCoolNowRight (approved):

anthony9187 (approved):

PR #496, opened 2024-03-27
fix: allow withEditItems to work with lazyScroll

joshhowenstine:
### Description

An issue was discovered where withEditItems does not integrate smoothly with lazyScroll. This update ensures that the scrolling functionality will pause until the initial item swap transition is complete, thereby preventing any conflict or unexpected behavior during additional logic execution.

[Ticket](https://ccp.sys.comcast.net/browse/OTTX-31307)

### Testing Instructions

1. Navigate to the withEditItems story.
2. Scroll through the row, it should operate as expected.
3. Press "Enter" to enable edit mode and move the current item. The functionality should remain seamless.
4. Enable the lazyScroll option in Storybook.
5. Repeat the above actions. After swapping items, the scrolling animation should now proceed as expected without issues.


ImCoolNowRight (approved):
Looks great!

erautenberg (approved):

PR #501, opened 2024-04-22
fix: Round radius token inverting

ctoddy:
## Description

Note: The original ticket is a combo of 1410 & 1409 but 1410 was closed as a duplicate with all information migrated to 1409 which is why the branch was originally named 1410. 

Created a getMaxRoundRadius() util to prevent the border-radius from inverting on multiple components. 

To recreate the issue navigate to the Button component in the open-source LUI storybook. In the "Component Style Theme Values" tab edit the radius to 150 and observe border-radius inversion.

## References

[LUI-1409](https://ccp.sys.comcast.net/browse/LUI-1409)

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/Checkbox/Checkbox.js
+++ b/packages/@lightningjs/ui-components/src/components/Checkbox/Checkbox.js
@@ -124,7 +124,7 @@ export default class Checkbox extends Base {
         // Compensating for the extra 2 pixels getRoundRect adds
         this.w - 2,
         this.h - 2,
-        getMaxRoundRadius(this.style.radius, this.w, this.h, 0),
+        getMaxRoundRadius(this.style.radius, this.w - 2, this.h - 2),
         this.style.strokeWidth,
         this.style.strokeColor,
         false
--- a/packages/@lightningjs/ui-components/src/components/Checkbox/Checkbox.js
+++ b/packages/@lightningjs/ui-components/src/components/Checkbox/Checkbox.js
@@ -96,16 +96,19 @@ export default class Checkbox extends Base {
       ? this.style.backgroundColorChecked
       : this.style.backgroundColor;
 
+    const width = this.w - this.style.strokeWidth * 2 - 2;
+    const height = this.h - this.style.strokeWidth * 2 - 2;
+  
     this._Body.patch({
       texture: lng.Tools.getRoundRect(
         // Compensating for the extra 2 pixels getRoundRect adds
-        this.w - this.style.strokeWidth * 2 - 2,
-        this.h - this.style.strokeWidth * 2 - 2,
+        width,
+        height,
         getMaxRoundRadius(
           this.style.radius,
-          this.w,
-          this.h,
-          this.style.strokeWidth - 2
+          width,
+          height,
+          this.style.strokeWidth * 2 - 2
         ),
         0,
         null,
--- a/packages/@lightningjs/ui-components/src/components/Checkbox/Checkbox.js
+++ b/packages/@lightningjs/ui-components/src/components/Checkbox/Checkbox.js
@@ -20,6 +20,7 @@ import lng from '@lightningjs/core';
 import Icon from '../Icon';
 import Base from '../Base';
 import * as styles from './Checkbox.styles';
+import { getMaxRoundRadius } from '../../utils';
 
 export default class Checkbox extends Base {
   static get __componentName() {
@@ -106,7 +107,7 @@ export default class Checkbox extends Base {
         // Compensating for the extra 2 pixels getRoundRect adds
         this.w - this.style.strokeWidth * 2 - 2,
         this.h - this.style.strokeWidth * 2 - 2,
-        radius,
+        getMaxRoundRadius(this.style.radius, this.w, this.h, this.style.strokeWidth - 2),
         0,
         null,
         true,
@@ -121,7 +122,7 @@ export default class Checkbox extends Base {
         // Compensating for the extra 2 pixels getRoundRect adds
         this.w - 2,
         this.h - 2,
-        this.style.radius,
+        getMaxRoundRadius(this.style.radius, this.w, this.h, 0),
         this.style.strokeWidth,
         this.style.strokeColor,
         false
--- a/packages/@lightningjs/ui-components/src/components/Checkbox/Checkbox.js
+++ b/packages/@lightningjs/ui-components/src/components/Checkbox/Checkbox.js
@@ -96,18 +96,17 @@ export default class Checkbox extends Base {
       ? this.style.backgroundColorChecked
       : this.style.backgroundColor;
 
-    // if the inner checkbox should be square, a rounded corner radius can still be applied to the stroke
-    const radius =
-      this.style.radius >= this.w / 2
-        ? (this.w - this.style.strokeWidth - 2) / 2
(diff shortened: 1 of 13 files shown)


packages/@lightningjs/ui-components/src/utils/index.js
```diff
@@ -854,6 +854,16 @@ export function convertTextAlignToFlexJustify(align) {
   }
 }
 
+/**
+ * Write params in here soon
+ * function is checking for single num or array
+ */
+export function getMaxRoundRadius(radius, width, height, offset){
+  const maxRadius = r => Math.min(radius, (Math.min(width, height) / 2)) + offset;
```
joshhowenstine: Change to const maxRadius = r => Math.min(r, (Math.min(width, height) / 2)) + offset;

joshhowenstine:
yarn lint:fix should fix any linter issues for you

packages/@lightningjs/ui-components/src/utils/index.js
```diff
@@ -854,6 +854,16 @@ export function convertTextAlignToFlexJustify(align) {
   }
 }
 
+/**
```
joshhowenstine, line 859: Finish writing comments for documentation

packages/@lightningjs/ui-components/src/utils/index.js
joshhowenstine: It also might be useful to cast to a number 

export function getMaxRoundRadius(radius, width, height, offset) {
  const maxRadius = r => Math.min(r, Math.min(width, height) / 2) + offset;
  return Array.isArray(radius)
    ? radius.map(r => Number(maxRadius(r)) || 0)
    : Number(maxRadius(radius)) || 0;
}

ImCoolNowRight (approved):
Looks great!

erautenberg (commented):
Not a huge deal, but you don't have to pass the "0" into the `getMaxRoundRadius` function, you could just leave it off and in the function, set the default of the offset to 0. Just one less thing to pass around!

packages/@lightningjs/ui-components/src/components/Shadow/Shadow.js
```diff
@@ -75,7 +77,7 @@ export default class Shadow extends Base {
               h: this.h - holepunchError,
               x: shadowSize + holepunchError / 2,
               y: shadowSize + holepunchError / 2,
-              radius: this.style.radius
+              radius: radius
```
erautenberg: you could simplify this to just `radius` since the key and variable have the same name

packages/@lightningjs/ui-components/src/utils/index.js
```diff
@@ -854,6 +854,24 @@ export function convertTextAlignToFlexJustify(align) {
...
+ * Prevents user input radius from inverting.
+ *
+ * @param {number} user input radius
+ * @param {number} width
+ * @param {number} height
+ * @param {number} offset
+ * @return {number} max radius for object before border begins to invert
+ */
+export function getMaxRoundRadius(radius, width, height, offset) {
+  //const maxRadius = r => Math.min(r, (Math.min(width, height) / 2) + offset);
```
erautenberg: remove?

packages/@lightningjs/ui-components/src/components/Checkbox/Checkbox.js
```diff
@@ -95,18 +96,17 @@ export default class Checkbox extends Base {
...
         // Compensating for the extra 2 pixels getRoundRect adds
         this.w - this.style.strokeWidth * 2 - 2,
         this.h - this.style.strokeWidth * 2 - 2,
-        radius,
+        getMaxRoundRadius(
+          this.style.radius,
+          this.w,
+          this.h,
+          this.style.strokeWidth - 2
+        ),
```
erautenberg, line 109: Not sure it's specific to this PR, it might be happening on develop, but in the Component Style Theme Values panel, if you increase the radius, only the border seems to change, the interior square does not.

packages/@lightningjs/ui-components/src/components/Radio/Radio.js
```diff
@@ -93,7 +88,7 @@ export default class Radio extends Base {
       texture: lng.Tools.getRoundRect(
         this.w - 2,
         this.h - 2,
-        this.style.radius,
+        getMaxRoundRadius(this.style.radius, this.w, this.h, 0),
```
erautenberg, line 91: same as the above comment about Checkbox. If I set the radius in the component styles to 15, I get odd behavior.

packages/@lightningjs/ui-components/src/components/Radio/Radio.js
```diff
@@ -67,18 +68,12 @@ export default class Radio extends Base {
...
-        ? (this.w - this.style.strokeWidth - 2) / 2
-        : 0;
-
     this._Body.patch({
       texture: lng.Tools.getRoundRect(
         // Compensating for the extra 2 pixels getRoundRect adds
         this.w - this.style.strokeWidth * 2 - 2,
         this.h - this.style.strokeWidth * 2 - 2,
-        radius,
+        getMaxRoundRadius(this.style.radius, this.w, this.h, 0),
```
erautenberg: I think the width and height here need to match the value above `this.w - this.style.strokeWidth * 2` and `this.h - this.style.strokeWidth * 2`

packages/@lightningjs/ui-components/src/components/Checkbox/Checkbox.js
```diff
@@ -95,18 +96,17 @@ export default class Checkbox extends Base {
...
       texture: lng.Tools.getRoundRect(
         // Compensating for the extra 2 pixels getRoundRect adds
         this.w - this.style.strokeWidth * 2 - 2,
         this.h - this.style.strokeWidth * 2 - 2,
-        radius,
+        getMaxRoundRadius(
+          this.style.radius,
+          this.w,
+          this.h,
+          this.style.strokeWidth - 2
```
erautenberg: I think the width and height here need to match the value above `this.w - this.style.strokeWidth * 2` and `this.h - this.style.strokeWidth * 2`

packages/@lightningjs/ui-components/src/components/Tooltip/Tooltip.js
```diff
@@ -93,7 +94,12 @@ export default class Tooltip extends Base {
           type: Bubble,
           w: this._Background.w,
           h: this._Background.h,
-          radius: this.style.radius,
+          radius: getMaxRoundRadius(
+            this.style.radius,
+            this._Background.w / 2,
+            this._Background.h / 2,
```
erautenberg: remove the divide by two for both the width and height, and the height should be `this._Background.h - this.style.pointerHeight` to exclude the pointer that is part of that entire background texture and just focus on the height of the bubble part. You can also remove the w/h setters on 91/92, they're redundant from `_updateBackgroundHeight`

erautenberg (changes requested):

ctoddy:
@erautenberg when I take the 0 out it breaks, the behavior I'm seeing is the radius not adjusting and console errors

erautenberg:
@ctoddy in the radius function you need to add a default, like this
```js
export function getMaxRoundRadius(radius, width, height, offset = 0)
```

erautenberg (approved):

PR #504, opened 2024-04-29
fix(Tab): tab padding properties match Button properties

ctoddy:
## Description

Tab component property ( formerly `noTitlePaddingX`, now `paddingXNoTitle` ) was updated and aliased to match the semantic Button property `paddingXNoTitle`. 

## References

[LUI-1420](https://ccp.sys.comcast.net/browse/LUI-1420)

## Checklist


packages/@lightningjs/ui-components/src/components/TabBar/Tab.d.ts
```diff
@@ -29,7 +29,7 @@ type TabStyle = SurfaceStyle & {
   contentColor: Color;
   iconMarginRight: number;
   iconSize: number;
-  noTitlePaddingX: number;
+  paddingXNoTitle: number;
```
erautenberg, line 38: keep the old one here too and add the `@deprecated` comment above it. Check out Button.d.ts for reference

packages/@lightningjs/ui-components/src/components/TabBar/Tab.d.ts
```diff
@@ -29,6 +29,7 @@ type TabStyle = SurfaceStyle & {
   contentColor: Color;
   iconMarginRight: number;
   iconSize: number;
+  /** @deprecated */
   noTitlePaddingX: number;
   paddingX: number;
```
ImCoolNowRight, line 39: You'll still need `paddingXNoTitle` here so that it is made available to TypeScript devs. The `/** @deprecated */` for `noTitlePaddingX` will let devs know that it's no longer supported, and the `aliasStyles` you set up in Tab.js will allow the old style to still function until devs update their code, but you'll need `paddingXNoTitle` in the TypeScript definition so that devs can use the new style you set up.

packages/@lightningjs/ui-components/src/components/TabBar/Tab.styles.js
```diff
@@ -20,6 +20,7 @@ export const base = theme => ({
   radius: theme.radius.xl,
   paddingX: theme.spacer.xxxl + theme.spacer.xxs,
   paddingY: theme.spacer.md + theme.spacer.xs,
+  /** @deprecated */
```
ImCoolNowRight: You don't need this in styles files, only TypeScript definition files. You 'll just update the `noTitlePaddingX` to `paddingXNoTitle`.

ImCoolNowRight (changes requested):

packages/@lightningjs/ui-components/src/components/TabBar/Tab.styles.js
ctoddy: Fixed

ImCoolNowRight (approved):

packages/@lightningjs/ui-components/src/components/TabBar/Tab.d.ts
anthony9187: regarding this, when possible we should add a message for some context around the deprecation - whether that's "this went away cause this" and/or "use x instead"

anthony9187 (approved):

PR #513, opened 2024-06-06
fix(ListItemSlider): value updates with min and max

ctoddy:
## Description

If the min exceeded the value OR the max was less than the value, the slider value itself was not updating until the arrows were used triggering an update in the value. 

Adding a check in _updateValue() allows the value to update with regard to its minimum and maximum properly. 

_how many times did I say **value** in this pr_

## References

LUI-1491

## Testing

- Run the branch locally
- Change the minimum to be greater than the value
- Observe the value automatically updating to be the minimum
- Repeat by making the value greater than the max and observe the max being set as the value
- Use the arrow keys to confirm the component operates by decreasing/increasing value when expected

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/ListItem/ListItemSlider.js
+++ b/packages/@lightningjs/ui-components/src/components/ListItem/ListItemSlider.js
@@ -92,6 +92,8 @@ export default class ListItemSlider extends ListItem {
 
   _updateValue() {
     if (this._hasValue) {
+      this._Slider.value = this.value =
+        this.value > this.max ? this.max : Math.max(this.min, this.value);
       let valuePatch = {
         content: this.value.toString(),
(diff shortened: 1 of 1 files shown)


erautenberg:
I'm not sure I agree with this fix. This sets the value of the Slider itself twice, first in the `_updateSliderPosition` method, then in `_updateValue`. I don't know how often this use case would come up, but this means if the user sets the value, then updates the minimum, even if the minimum is still less than the current value, the current value gets replaced with the new min.

https://github.com/rdkcentral/Lightning-UI-Components/assets/15019089/bb5b2e13-9c0a-496e-aa96-ffd71e64e18e

ctoddy:
Closing because opened https://github.com/rdkcentral/Lightning-UI-Components/pull/515 in a different branch

PR #515, opened 2024-06-11
fix(ListItemSlider): values sync with SB controls and textbox

ctoddy:
## Description

The ListItemSlider's component value was not syncing between the component's actual textbox value (when you adjust the value with L/R arrows) and the input storybook control value. By using the storybook client API and signals the storybook value is now responsive (staying in sync) with the value of the slider itself. Similar logic is already used within the Slider component to keep its values synced. 

This fix solves the issue of the min and max values not being updated until the arrows were used.

## References

LUI-1491

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/ListItem/ListItemSlider.js
+++ b/packages/@lightningjs/ui-components/src/components/ListItem/ListItemSlider.js
@@ -80,6 +80,12 @@ export default class ListItemSlider extends ListItem {
     super._update();
     this._updateSliderPosition();
     this._updateValue();
+    // Added this below
+    if (this._valueChanged) {
+      this.signal('onChange', this.value, this);
+      this.fireAncestors('$announce', this.announce);
+      this._valueChanged = false;
+    }
   }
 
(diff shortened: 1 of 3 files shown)


erautenberg (commented):
The value issue is definitely resolved, but the actual slider bar doesn't seem to update correctly. For example, if starting with a range of 0 to 100 and a value of 75, the slider is, as expected, 3/4 of the way full. However, if you then change the minimum value to 50, 75 should be halfway between 50 and 100. I think there is another step missing to update the progress amount of the bar itself. Looks like line 190 isn't correct as it's not including the absolute value of the max minus the min.

ctoddy:
@erautenberg would this be considered a separate issue/ticket from the original issue? This is how the Slider is currently operating apart from the ticket.

erautenberg:
@ctoddy that's fine with me, please work with Nassar to make a new ticket to track! thank you

joshhowenstine (approved):

ImCoolNowRight (approved):

PR #517, opened 2024-06-12
fix(FocusManager): add zIndex override ability

ImCoolNowRight:
## Description

Adds the ability to add a `selectedZindex` and `defaultZIndex` for the selected item and non-selected items, respectively.

## References

LUI-1341

## Testing

- Run `yarn build` and copy the `dist` folder into the `@lightningjs/ui-components` directory in the IS `node_modules`.
- Set `this._selectedZIndex = 2;` in `_construct()` for Galery3x4CoverArt, then confirm that the selected Tile in the Gallery3x4CoverArt story is always on top of any non-selected item as you navigate around the gallery. (This will require an IS PR to make this change once this is confirmed to work properly.)

## Automation

N/A

## Checklist


anthony9187 (approved):

ChrisArasin:
I wonder if this needs to be an opt-in feature. Like if you don't have a default or selectedZindex set it doesn't run? Otherwise it will be a sneaky breaking change for anything manipulating zIndex inside a focus manager.

anthony9187:
> I wonder if this needs to be an opt-in feature. Like if you don't have a default or selectedZindex set it doesn't run? Otherwise it will be a sneaky breaking change for anything manipulating zIndex inside a focus manager.

This is a really good point, and also has me wondering where this came from now

slhay28 (approved):

PR #521, opened 2024-06-25
fix(Textbox): stops flash on inlineContent change

slhay28:
## Description

It was noticed that the TextBox would flash when loading. This would occur when the TextBox contained content that were of type `InlineContent` and any property changed. This was due to the number of lines in the flex container potentially being more than the maxLines within the `InlineContent`. It would then calculate the height based on this wrong number and send it back to the TextBox and flash the wrong number of lines. 

## References

LUI-1435

## Testing

- Check to make sure everything within both the `InlineContent` stories and `TextBox` stories renders as expected.
- Specifically, make sure that when changing the `maxLines` control, it works as expected and that there is no flashing of the component.

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
+++ b/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
@@ -147,30 +147,34 @@ export default class InlineContent extends Base {
     // perhaps have to wait until Lightning Flexboxes can emit a signal (like textures) when they've finished loading
     if (this.children.length) {
       setTimeout(() => {
+        debugger
+        this.stage.update()
         this.multiLineHeight = this.finalH;
+
         if (
           this.flex &&
           this.flex._layout &&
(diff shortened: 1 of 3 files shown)


ctoddy:
Screenshot 2024-06-25 at 11 11 36 AM
Not sure why but every time I load the component it shows up cut off like the screenshot here. Not sure if that's expected. I'm able to get around it by adjusting the justifyContent prop.

ctoddy:
https://github.com/rdkcentral/Lightning-UI-Components/assets/32280146/f567b34a-2ce5-40f7-82a6-34c047d5a0da

I'm somewhat able to still see the flashing when I switch between wrapped and not wrapped, but I'm not sure how realistic this scenario would be for a user/ seems like this could be outside of the scope of the original problem you fixed.

slhay28:
@ctoddy 
- regarding the component showing up cut off- I believe there is another ticket regarding that and it is not a result of this pr
- regarding the flash, ill look back into that

PR #532, opened 2024-09-10
fix(metadataCardContent): Provider covers text after decrementing value

shanteaustin:
## Description

It has been observed in Storybook on the metadataCardContent Component, that when a provider count is increased to max 9 value and then decreased to a single provider, the effects are still masked over the text. 

## Testing

1. Go to metadataCardContent 
2. Change number of visible providers to 9 
3. When number of visible providers is 9 Detail text should be masked
4. Change number of visible providers to 1
5. When number of visible providers is 1 Detail text should not be masked. 

## Checklist

- [X] all commented code has been removed
- [X] any new console issues have been resolved
- [X] code linter and formatter has been run
- [X] test coverage meets repo requirements
- [X] PR name matches the expected semantic-commit syntax

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/MetadataCardContent/MetadataCardContent.js
+++ b/packages/@lightningjs/ui-components/src/components/MetadataCardContent/MetadataCardContent.js
@@ -21,7 +21,6 @@ import TextBox from '../TextBox';
 import Provider from '../Provider';
 import FadeShader from '../../shaders/FadeShader';
 import * as styles from './MetadataCardContent.styles';
-import debounce from 'debounce';
 
 export default class MetadataCardContent extends MetadataBase {
   static get __componentName() {
--- a/packages/@lightningjs/ui-components/src/components/MetadataCardContent/MetadataCardContent.js
+++ b/packages/@lightningjs/ui-components/src/components/MetadataCardContent/MetadataCardContent.js
@@ -108,11 +108,6 @@ export default class MetadataCardContent extends MetadataBase {
     ];
   }
 
-  _init() {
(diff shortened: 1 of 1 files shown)


cee-money (approved):
LGTM 🚀

packages/@lightningjs/ui-components/src/components/MetadataCardContent/MetadataCardContent.js
```diff
@@ -233,6 +233,7 @@ export default class MetadataCardContent extends MetadataBase {
 
     this._Provider.x = this._DetailsWrapper.w - this._providerW;
     this._Provider.y = this._DetailsWrapper.h - this._providerH;
+    this._updateDetails();
```
ImCoolNowRight: This feels weird to have at the end of `_updatePositions`. Can you try moving this to the end of `_resolveProvider` and see if this still does the trick? I tested and it seems like it might work. That will make it so that it updates that portion of the component after the Provider is done loading, which I think is more likely what we want instead of it running every time we need to update positions (which gets called by several other update functions).

cee-money: good callout!

shanteaustin: I'll take a look to make sure it's working as expected. Thanks @ImCoolNowRight

shanteaustin: Looks good. Made the update.

anthony9187 (approved):

PR #534, opened 2024-09-12
fix: make sure subThemes are taken into account when generating style cache

joshhowenstine:
We have identified a problem where the subThemes are not functioning as expected due to improper caching. This issue is affecting the normal operation of the subThemes feature and preventing it from performing as intended.

To resolve this, we are implementing a fix to properly manage the cache related to subThemes. Once this fix is applied, the subThemes should operate correctly, restoring full functionality and ensuring that all related features work as designed.


ImCoolNowRight (approved):
Assuming this does the trick for Ben, looks good!

cee-money (approved):
👍

PR #535, opened 2024-09-12
feat(InlineContent): Font Strikethrough

shanteaustin:
## Description

Updated text styles to include an option to strike through 

## Testing

Go to Inline Content --> With Strike Through 
To view line through text

## Checklist

- [X] all commented code has been removed
- [X] any new console issues have been resolved
- [X] code linter and formatter has been run
- [X] test coverage meets repo requirements
- [X] PR name matches the expected semantic-commit syntax

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
+++ b/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
@@ -89,8 +89,6 @@ export default class InlineContent extends Base {
         const base = {
           flexItem: {
             ...this.contentProperties,
-            // apply contentProperties object props first if those are defined
-            // otherwise will use the style props
             marginBottom: isLast ? 0 : this._marginBottom,
             marginRight: isLast
               ? 0
@@ -101,7 +99,14 @@ export default class InlineContent extends Base {
         // text not separated by icons/badges are grouped together
         if (isText(item)) {
           const nextItem = this._parsedContent[index + 1];
-          if (nextItem && isText(nextItem)) {
+          if (
+            (nextItem && isText(nextItem)) ||
+            (this.contentWrap &&
+              nextItem &&
+              nextItem.newline &&
+              this._parsedContent[index + 2] &&
+              isText(this._parsedContent[index + 2]))
+          ) {
             base.flexItem.marginRight = 0;
           }
           this.childList.a(this._createText(base, item));
@@ -368,8 +373,8 @@ export default class InlineContent extends Base {
       const strikethroughLine = {
         rect: true,
         w: textWidth,
-        color: textOverrideStyles?.textColor || 0xffffffff,
-        h: textComponent.h * 0.08,
+        color: textOverrideStyles?.textColor || this.style.textDefaultColor,
+        h: textComponent.h * this.style.strikethroughRatio,
         y: textComponent.h / 2,
         mountY: 1
       };
--- a/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
+++ b/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
@@ -367,7 +367,7 @@ export default class InlineContent extends Base {
       });
       const strikethroughLine = {
         rect: true,
-        w: textWidth - 2,
+        w: textWidth,
         color: textOverrideStyles?.textColor || 0xffffffff,
         h: textComponent.h * 0.08,
         y: textComponent.h / 2,
--- a/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
+++ b/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
@@ -15,7 +15,7 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  */
-
+import lng from '@lightningjs/core';
 import Icon from '../Icon';
 import Badge from '../Badge';
 import { parseInlineContent, flatten, measureTextWidth } from '../../utils';
@@ -346,8 +346,7 @@ export default class InlineContent extends Base {
       typeof text.style === 'string'
         ? this.customStyleMappings[text.style]
         : text.style;
-
-    return {
+    const textComponent = {
       ...base,
       y: this.textY !== undefined ? this.textY : this.style.textY,
       h:
@@ -360,6 +359,26 @@ export default class InlineContent extends Base {
         text: text.text || text
       }
     };
+    if (textOverrideStyles?.textDecoration === 'line-through') {
+      const textWidth = measureTextWidth({
+        ...this.style.textStyle,
+        ...textOverrideStyles,
+        text: text.text || text
+      });
+      const strikethroughLine = {
+        rect: true,
+        w: textWidth - 2,
+        h: 2,
+        color: textOverrideStyles?.textColor || 0xffffffff,
+        y: textComponent.h / 2
+      };
+      return {
+        type: lng.Component,
+        w: textWidth + 5,
+        children: [{ ...textComponent }, { ...strikethroughLine }]
+      };
+    }
+    return textComponent;
   }
 
   _createBadge(base, badge) {
--- a/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
+++ b/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
@@ -375,7 +375,7 @@ export default class InlineContent extends Base {
       };
       return {
         type: lng.Component,
-        w: textWidth,
+        w: textWidth + textComponent.flexItem.marginRight,
         h: textComponent.h,
         children: [{ ...textComponent }, { ...strikethroughLine }]
       };
--- a/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
+++ b/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
@@ -373,7 +373,10 @@ export default class InlineContent extends Base {
       const strikethroughLine = {
         rect: true,
         w: textWidth,
-        color: textOverrideStyles?.textColor || this.style.textDefaultColor,
+        color:
+          this.style.strikethroughColor ||
+          textOverrideStyles?.textColor ||
+          this.style.textStyle.textColor,
         h: textComponent.h * this.style.strikethroughRatio,
         y: textComponent.h / 2,
         mountY: 1
--- a/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
+++ b/packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
@@ -368,13 +368,15 @@ export default class InlineContent extends Base {
       const strikethroughLine = {
         rect: true,
         w: textWidth - 2,
-        h: 2,
         color: textOverrideStyles?.textColor || 0xffffffff,
-        y: textComponent.h / 2
+        h: textComponent.h * 0.08,
+        y: textComponent.h / 2,
+        mountY: 1
       };
       return {
         type: lng.Component,
-        w: textWidth + 5,
+        w: textWidth,
+        h: textComponent.h,
         children: [{ ...textComponent }, { ...strikethroughLine }]
       };
     }
@@ -527,6 +529,9 @@ export default class InlineContent extends Base {
           announce += item.announce;
         } else if (item.text) {
           announce += item.text;
+          if (item.style?.textDecoration === 'line-through') {
+            announce += 'strikethrough';
+          }
         } else if (item.title) {
           announce += item.title;
         } else if (item.badge) {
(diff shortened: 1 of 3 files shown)


packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.stories.js
```diff
@@ -231,3 +231,37 @@ WithTruncation.args = {
...
+          w: 400,
+          contentWrap: args.contentWrap,
+          contentProperties: args.contentProperties,
+          justify: args.justify,
+          customStyleMappings: {
+            italic: {
+              fontStyle: 'italic',
+              textColor: getHexColor('FF6194')
+            }
+          },
```
ImCoolNowRight: This can be removed since this style isn't being used in this story.

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
```diff
@@ -360,6 +359,26 @@ export default class InlineContent extends Base {
...
+        w: textWidth - 2,
+        h: 2,
+        color: textOverrideStyles?.textColor || 0xffffffff,
+        y: textComponent.h / 2
+      };
+      return {
+        type: lng.Component,
+        w: textWidth + 5,
+        children: [{ ...textComponent }, { ...strikethroughLine }]
+      };
```
ImCoolNowRight: I think we're going to have to workshop this. It seems to be causing the text after it to overlap when `contentWrap` is true.

Screenshot 2024-09-12 at 4 02 49 PM

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.stories.js
shanteaustin: Removed

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
```diff
@@ -360,6 +359,26 @@ export default class InlineContent extends Base {
...
+    if (textOverrideStyles?.textDecoration === 'line-through') {
+      const textWidth = measureTextWidth({
+        ...this.style.textStyle,
+        ...textOverrideStyles,
+        text: text.text || text
+      });
+      const strikethroughLine = {
+        rect: true,
+        w: textWidth - 2,
+        h: 2,
```
cee-money: the `h` might need to be dynamic (a percentage of the text component's h). for example what if the font size is 40? the strikethrough might appear too thin.

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
```diff
@@ -360,6 +359,26 @@ export default class InlineContent extends Base {
...
+        color: textOverrideStyles?.textColor || 0xffffffff,
+        y: textComponent.h / 2
```
cee-money: we may want to add `mountY: 0.5` so that the strikethrough is fully vertically centered in the text

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
shanteaustin: Great point

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
shanteaustin: The mountY doesnt work on my end to center. Did you test this out?

cee-money: i did, it's definitely minor 😆

but here's what i'm seeing. `mountY: 1` might be best? 
Screenshot 2024-09-13 at 11 23 58 AM

cee-money: or potentially something like
```
h: textComponent.h * 0.08,
y: textComponent.h / 2 - textComponent.h * 0.08,
mountY: 0.5
```

ImCoolNowRight: Just to throw my nitpicking two cents in, I think this looks best:
```js
h: textComponent.h * 0.08,
y: textComponent.h / 2,
mountY: 1
```

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
```diff
@@ -360,6 +359,26 @@ export default class InlineContent extends Base {
...
+      };
+      return {
+        type: lng.Component,
+        w: textWidth + 5,
```
ImCoolNowRight: Adding `h: textComponent.h` here will resolve the overlapping issue when contentWrap is true.

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
ImCoolNowRight, line 1: We're going to have to update the announce method to accommodate for the strikethrough. I posed the question to the team about whether "strikethrough" should come before or after the text it is applied to, but essentially, this will need to be update in `announce()`:

```js
else if (item.text) {
  announce += item.text;

  if (item.style?.textDecoration === 'line-through') {
    announce += 'strikethrough';
  }
}
```

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
shanteaustin: Looks perfectly centered now. Thanks guys!

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.stories.js
```diff
@@ -231,3 +231,31 @@ WithTruncation.args = {
...
+          x: 550,
+          w: 400,
+          contentWrap: args.contentWrap,
+          contentProperties: args.contentProperties,
+          justify: args.justify,
+          content: [
+            'Rent',
+            { newline: true },
+            {
+              text: '$19.99 ',
```
cee-money: i think a trailing space sneaked in here

```suggestion
              text: '$19.99',
```

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
```diff
@@ -508,6 +529,9 @@ export default class InlineContent extends Base {
           announce += item.announce;
         } else if (item.text) {
           announce += item.text;
+          if (item.style?.textDecoration === 'line-through') {
+            announce += 'strikethrough';
+          }
```
cee-money, line 542: just checking, is override-able? @ImCoolNowRight i vaguely remember this coming up during our last sync. did we decide that users should be able to pass whatever announce text they want into this like: "Price used to be __ but is now __"?

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
```diff
@@ -360,6 +359,28 @@ export default class InlineContent extends Base {
...
+      const textWidth = measureTextWidth({
+        ...this.style.textStyle,
+        ...textOverrideStyles,
+        text: text.text || text
+      });
+      const strikethroughLine = {
+        rect: true,
+        w: textWidth - 2,
+        color: textOverrideStyles?.textColor || 0xffffffff,
+        h: textComponent.h * 0.08,
```
anthony9187: 1. the width and the height are both being calculated with some hardcoded values, are we cool with that?
2. should the fallback color be derived from a theme value?

ImCoolNowRight: 1. @shanteaustin Do we need the 2 being subtracted? I think the height is a good callout, we can pull the 0.08 out into a style.
2. I don't think it hurts to point to a token by default. At least the strikethrough would be in line with the user's theme.

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
ImCoolNowRight, line 1: The if statement on 104 should be updated to:

```js
if (
  (nextItem && isText(nextItem)) ||
  (this.contentWrap &&
    nextItem &&
    nextItem.newline &&
    this._parsedContent[index + 2] &&
    isText(this._parsedContent[index + 2]))
  ) {
```

As it stands right now, the spacing is off on some lines if you set `contentWrap` to true. This, along with the other comment in this review, will make sure that everything is spaced out properly.

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
```diff
@@ -360,6 +359,28 @@ export default class InlineContent extends Base {
...
+        y: textComponent.h / 2,
+        mountY: 1
+      };
+      return {
+        type: lng.Component,
+        w: textWidth,
```
ImCoolNowRight: This needs to be updated to `w: textWidth + textComponent.flexItem.marginRight,`

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
ImCoolNowRight: Still waiting on confirmation from the accessibility team. Personally, I think this is a good generic default, and that clients can override `_announce` for their needs on a case by case basis. But we'll see what they say!

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.stories.js
shanteaustin: If the space is not added the text is too close together. If you take a look at the file some of the text have spaces added.

shanteaustin: But I will remove it and push up. It may look good to you without the added space.

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
shanteaustin: 1. @ImCoolNowRight The line seemed at bit long to me but after a couple tweaks it's not so bad so I will remove that. When you say pull the 0.08 into a style do you mean adding it to the theme object in InlineContent.styles? 
2. Which token should we use? textNeutral?

ImCoolNowRight: 1. We have a `.styles.js` file for each component that has the styles used by it. We're suggesting adding a new style property to it, something like `strikethroughRatio`, that would be set to 0.08 by default, and then we reference that style instead of the hard coded number. That way, users can pass something else in if they want to.
2. I think textNeutral works!

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
shanteaustin: Will update as suggested.

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.stories.js
ImCoolNowRight: I think this space is the only thing left to change.

ImCoolNowRight (commented):
Just need to remove that extra space in the story, and then I think this will be good to go! I'm ok with circling back once we hear back about how to handle the announcer.

shanteaustin:
> Just need to remove that extra space in the story, and then I think this will be good to go! I'm ok with circling back once we hear back about how to handle the announcer.

I thought that was remove looks like it was reverted so I updated it again.

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
```diff
@@ -360,6 +364,28 @@ export default class InlineContent extends Base {
...
+    if (textOverrideStyles?.textDecoration === 'line-through') {
+      const textWidth = measureTextWidth({
+        ...this.style.textStyle,
+        ...textOverrideStyles,
+        text: text.text || text
+      });
+      const strikethroughLine = {
+        rect: true,
+        w: textWidth,
+        color: textOverrideStyles?.textColor || this.style.textDefaultColor,
```
ImCoolNowRight: Sorry, one last thing!

Can we actually reverse this order and rename the style property to `strikethroughColor` (and update the style file, too)? I think that way makes more sense so that users can define a color they want the strikethrough to be, and then we provide fallbacks.

So I think this should be `color: this.style.strikethroughColor || textOverrideStyles?.textColor || this.style.textStyle.textColor`. That'll fall back to the text override first, and then the default text style.

shanteaustin: Sounds good. I just updated the property name and changed the order.

ImCoolNowRight (approved):

cee-money (approved):
LGTM

packages/@lightningjs/ui-components/src/components/InlineContent/InlineContent.js
ImCoolNowRight: Resolving this conversation to get this merged in. Will create a follow-up ticket to ensure this gets completed.

PR #543, opened 2024-10-04
fix(Colors): enable [token, newAlpha] format

ImCoolNowRight:
## Description

Enables users to set colors using the format `[colorToken, newAlpha]` so that an existing color token can be utilized and just have its alpha value overridden.

This was doable prior to #499, but was only used in a few fringe cases. This PR enables that again while maintaining the functionality added in the previous PR.

## References

LUI-1592

## Testing

Set the `backgroundColor` in Surface.styles.js to be `[theme.color.interactiveNeutral, theme.alpha.alpha1]`. This should no longer cause the component to render in pure white. Change the second value to `theme.alpha.alpha2` and ensure it adjusts accordingly.

## Checklist

### DIFF
--- a/packages/@lightningjs/ui-components/src/mixins/withThemeStyles/utils.js
+++ b/packages/@lightningjs/ui-components/src/mixins/withThemeStyles/utils.js
@@ -557,19 +557,30 @@ export const colorParser = (targetObject, styleObj) => {
   // Process style object and remove unnecessary properties
   const processedStyle = JSON.stringify(styleObj, (_, value) => {
     if (-1 < ['tone', 'mode'].indexOf(_)) return value; // Remove any tone/mode or mode/tone properties as they have already been processed
-    if ('string' === typeof value && value.startsWith('theme.')) {
-      // Support theme strings example: theme.radius.md
-      return getValFromObjPath(targetObject, value); // If no theme value exists, the property will be removed from the object
-    } else if (
+
+    // Handle theme strings, e.g., 'theme.radius.md'
+    if (typeof value === 'string' && value.startsWith('theme.')) {
+      // Retrieve the value from the target object using the theme path
+      return getValFromObjPath(targetObject, value); // If no theme value exists, the property will be removed
+    }
+
+    function isValidColor(num) {
+      return num >= 0 && num <= 0xffffffff;
+    }
+
+    // Handle color arrays, e.g., ['#663399', 1] or [255, 0.5]
+    if (
       Array.isArray(value) &&
       value.length === 2 &&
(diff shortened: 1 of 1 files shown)


ImCoolNowRight:
@erautenberg Regarding [this change](https://github.com/rdkcentral/Lightning-UI-Components/pull/499/files#diff-63ccff1d65ce2d205ca47322f676f9890010560e78ce4ae5a7f9894475e21acbR392-R394) from #499 and whether it should be adjusted in this PR, I just want to make sure that you agree that we should only be allowing this format to work within style files and that there's no expectation of this ever coming from the theme and we don't need to adjust those conditionals, correct?

packages/@lightningjs/ui-components/src/mixins/withThemeStyles/utils.js
```diff
@@ -557,19 +557,30 @@ export const colorParser = (targetObject, styleObj) => {
...
-      return getValFromObjPath(targetObject, value); // If no theme value exists, the property will be removed from the object
-    } else if (
+
+    // Handle theme strings, e.g., 'theme.radius.md'
+    if (typeof value === 'string' && value.startsWith('theme.')) {
+      // Retrieve the value from the target object using the theme path
+      return getValFromObjPath(targetObject, value); // If no theme value exists, the property will be removed
+    }
+
+    function isValidColor(num) {
```
joshhowenstine, line 567: Should we move this into a common util or just use it inline since it is only used in one place?

joshhowenstine (approved):

packages/@lightningjs/ui-components/src/mixins/withThemeStyles/utils.js
ImCoolNowRight: I think it's fine here for now, since it's only for this one check for this one specific scenario we're adding consideration for. Obviously open to moving it in the future if it's needed elsewhere!

cee-money (approved):
LGTM

erautenberg:
> @erautenberg Regarding [this change](https://github.com/rdkcentral/Lightning-UI-Components/pull/499/files#diff-63ccff1d65ce2d205ca47322f676f9890010560e78ce4ae5a7f9894475e21acbR392-R394) from #499 and whether it should be adjusted in this PR, I just want to make sure that you agree that we should only be allowing this format to work within style files and that there's no expectation of this ever coming from the theme and we don't need to adjust those conditionals, correct?

Yep, I think that makes total sense. In the context of the original theme, you could always reference just the first value of another token instead of the entire thing, so there are clear workarounds from that level.

PR #544, opened 2024-10-16
feat: Add `withTextMagnifier` mixin for enabling magnified text on focus

joshhowenstine:
This PR introduces a new mixin called `withTextMagnifier` that can be applied to lightning applications b. The mixin allows a `TextMagnifier` component to dynamically display magnified text when focus changes occur within the application. 

Oct-18-2024 12-23-44

The key functionality includes:
- Enabling or disabling the text magnifier with the `textMagnifierEnabled` property.
- Automatically detecting focused elements and dynamically updating the `TextMagnifier` based on the focus text (title, description, or announce).
- Displaying the magnifier either at the top or bottom of the screen, depending on the location of the focused element.
  
`TextMagnifier` component itself has been created and includes a `ScrollWrapper`, which manages the scrolling behavior of long content. Position, radius, and size, are automatically adjusted based on the component's location on the screen as well as theme values

#### References:
- The `TextMagnifier` component is styled according to the specifications in `TextMagnifier.styles.js`.
- Uses the `ScrollWrapper` component for auto-scrolling within the magnified text view.

#### Testing:
- Verified the correct instantiation and rendering of the `TextMagnifier` component when `textMagnifierEnabled` is set to `true`.
- Tested various focus scenarios, including focus transitions between different elements with `title`, `description`, or `announce` properties.
- Confirmed correct positioning of the `TextMagnifier` based on the location of the focused element (top or bottom of the screen).
- Tested scenarios where the `TextMagnifier` is disabled or focus text is empty to ensure the magnifier is not displayed.
- Respects neutral, inverse, and brand tone when set on the component

#### Automation:
- Basic unit tests

#### Checklist:
- [x] All commented code has been removed.
- [x] Any new console issues have been resolved.
- [x] Code linter and formatter have been run and passed.
- [ ] Test coverage meets the repository's requirements.
- [x] PR name matches the expected semantic-commit syntax.

### DIFF
--- a/packages/@lightningjs/ui-components/src/components/TextMagnifier/TextMagnifier.d.ts
+++ b/packages/@lightningjs/ui-components/src/components/TextMagnifier/TextMagnifier.d.ts
@@ -1,0 +1,20 @@
+import Surface, { SurfaceStyle } from '../Surface';
+import { StylePartial } from '../../types/lui';
+
+type TextMagnifierStyle = SurfaceStyle & {
+  color: string;
+  gutterX: number;
+  gutterY: number;
+  h: number;
+  marginX: number;
+  radius: number | string;
+  textStyle: any;
+};
+
+export default class TextMagnifier extends Surface {
+  get style(): TextMagnifierStyle;
+  set style(v: StylePartial<TextMagnifierStyle>);
+  location: 'top' | 'bottom';
+  mode: string;
+  content: string | null;
+}
--- a/packages/@lightningjs/ui-components/src/components/TextMagnifier/TextMagnifier.stories.js
+++ b/packages/@lightningjs/ui-components/src/components/TextMagnifier/TextMagnifier.stories.js
@@ -1,0 +1,43 @@
+/**
+ * Copyright 2023 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import lng from '@lightningjs/core';
+import { default as TextMagnifierComponent } from '.';
+import { context } from '@lightningjs/ui-components/src';
+
+export default {
+  title: 'Components/TextMagnifier'
+};
+
+export const TextMagnifier = () =>
+  class TextMagnifier extends lng.Component {
+    static _template() {
+      return {
+        TextMagnifier: {
+          type: TextMagnifierComponent
+        }
+      };
+    }
+
+    _init() {
+      this.parent.x = -context.theme.layout.marginX;
+      this.parent.y = -context.theme.layout.marginY;
+
+      super._init();
+    }
+  };
--- a/packages/@lightningjs/ui-components/src/components/TextMagnifier/TextMagnifier.stories.js
+++ b/packages/@lightningjs/ui-components/src/components/TextMagnifier/TextMagnifier.stories.js
@@ -30,7 +30,8 @@ export const TextMagnifier = () =>
       return {
         TextMagnifier: {
           type: TextMagnifierComponent,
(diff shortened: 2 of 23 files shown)


packages/@lightningjs/ui-components/src/components/TextMagnifier/TextMagnifier.stories.js
```diff
@@ -0,0 +1,45 @@
...
+
+import lng from '@lightningjs/core';
+import { default as TextMagnifierComponent } from '.';
+import { context } from '@lightningjs/ui-components/src';
+
+export default {
+  title: 'Components/TextMagnifier'
+};
+
+export const TextMagnifier = () =>
```
ImCoolNowRight, line 27: Can you change this const to "Basic" to match our other stories? Doing this seems to not nest it in a folder like the other components.

packages/@lightningjs/ui-components/src/components/TextMagnifier/TextMagnifier.d.ts
```diff
@@ -0,0 +1,20 @@
...
+import { StylePartial } from '../../types/lui';
+
+type TextMagnifierStyle = SurfaceStyle & {
+  color: string;
+  gutterX: number;
+  gutterY: number;
+  h: number;
+  marginX: number;
+  radius: number | string;
+  textStyle: any;
```
ImCoolNowRight, line 11: Should this be set to `TextBoxStyle`, or maybe `lng.textures.TextTexture.Settings | string` (which `textStyle` is set to in TextBox.style.d.ts)?

ImCoolNowRight:
Should the ScrollWrapper restart from the top after a set period of time after hitting the end of the content? Right now, we hit the end and just stay there, but I can imagine users expecting it to restart.

packages/apps/lightning-ui-docs/.storybook/addons/toolbars/Announce.js
```diff
@@ -50,3 +50,34 @@ export const Announce = memo(function MyAddonSelector() {
...
+      actionName: 'Magnifier',
+      action: toggleMagnifier
+    });
+  }, [toggleMagnifier, api]);
+
+  return (
+    <IconButton
+      key={MAGNIFIER_ID}
+      active={isActive}
+      title="Toggle a11y announcing (voice guidance) of components"
```
ImCoolNowRight, line 77: This needs to be updated to a proper description.

packages/@lightningjs/ui-components/src/mixins/withTextMagnifier/withTextMagnifier.stories.js
ImCoolNowRight, line 1: Not the end of the world, and probably a fringe case, but can we have this story disregard the global TextMagnifier flag? I noticed that if you are on the story, toggling the top control on has no effect, but then you can turn it off. At least for this story, it feels like we should just keep the magnifier on regardless of the flag? I don't feel too strongly, and again, it's probably a fringe case, but something I noticed.

packages/@lightningjs/ui-components/src/mixins/withTextMagnifier/index.js
```diff
@@ -0,0 +1,78 @@
...
+      ];
+    }
+
+    get _location() {
+      const { py: focusedY } = this._focusedElement?.core?.renderContext || {};
+      return focusedY > this.stage.h / 2 ? 'top' : 'bottom';
+    }
+
+    get _focusText() {
+      const { title, description, announce } = this._focusedElement || {};
```
ImCoolNowRight, line 45: I know this is endlessly expandable, so do we maybe want to make it a property users can define strings for and we can search for those on the focused element? Maybe this is the default? I'm just thinking that TextBox and InlineContent both have `content`, which I know `announce` would capture, but what if the user overrode that for whatever reason? I think offering the flexibility could come in handy.

packages/@lightningjs/ui-components/src/components/TextMagnifier/TextMagnifier.d.ts
```diff
@@ -0,0 +1,20 @@
+import Surface, { SurfaceStyle } from '../Surface';
+import { StylePartial } from '../../types/lui';
+
+type TextMagnifierStyle = SurfaceStyle & {
+  color: string;
+  gutterX: number;
+  gutterY: number;
+  h: number;
```
ImCoolNowRight, line 8: Is this not available by default?

packages/@lightningjs/ui-components/src/components/TextMagnifier/TextMagnifier.d.ts
```diff
@@ -0,0 +1,20 @@
...
+  marginX: number;
```
ImCoolNowRight, line 9: Is this being used anywhere?

ImCoolNowRight (changes requested):

packages/@lightningjs/ui-components/src/components/TextMagnifier/TextMagnifier.d.ts
joshhowenstine: Nope! Removing

packages/@lightningjs/ui-components/src/components/TextMagnifier/TextMagnifier.d.ts
joshhowenstine: Updated

packages/@lightningjs/ui-components/src/components/TextMagnifier/TextMagnifier.d.ts
joshhowenstine: Yup :)

packages/@lightningjs/ui-components/src/mixins/withTextMagnifier/index.js
joshhowenstine: You can now set a focusText method on the application and it will use that. What do you think?

packages/@lightningjs/ui-components/src/mixins/withTextMagnifier/withTextMagnifier.stories.js
joshhowenstine: Great idea. Added

PR #545, opened 2024-10-21
feat: add long press mixin

che-effe:
## Description

Creates a Long Press handler mixin that can work with any key. Component with this mixin has to set a threshold in ms to define long press length and also have a signal listers for `longPressHit` signal.
Handler method supplied will be called when long press threshold is reached and will recieve the key value that was pressed.
Client can then execute as they wish on any long pressed key they desire.

## References

LUI-60

## Testing

View `withLongPress` story in storybook and make sure the component works as described above.

## Checklist


packages/@lightningjs/ui-components/src/mixins/withLongPress/index.js
```diff
@@ -0,0 +1,79 @@
...
+      if (
+        // eslint-disable-next-line no-constant-condition
+        this.pressedTimeStart &&
+        keyEvent.timeStamp - this.pressedTimeStart > (this._threshold || 2000)
+      ) {
+        // shortcircuit here if we only want to execute once before a key up event
+        if (this._executeOnce && this.hasExecuted) {
+          return;
+        }
+        this.signal('longPressHit', keyEvent.key);
```
ImCoolNowRight: Would it make sense to change this to a `fireAncestor` call? I can see cases where this would be handled by a grandparent component, but I am also not against keeping this and saying that this is the preferred pattern.

packages/@lightningjs/ui-components/src/mixins/withLongPress/index.js
```diff
@@ -0,0 +1,79 @@
...
+
+    set targetKey(val) {
+      this._targetKey = val;
+    }
+    set executeOnce(val) {
+      this._executeOnce = val;
+    }
+    _init() {
+      this.pressedTimeStart = null;
+      this.hasExecuted = false;
```
ImCoolNowRight: Can we make these underscored to help indicate they're just for us and not for users?

packages/@lightningjs/ui-components/src/mixins/withLongPress/withLongPress.mdx
```diff
@@ -0,0 +1,72 @@
...
+
+import { Meta } from '@storybook/blocks';
+import * as withLongPressStories from './withLongPress.stories';
+
+<Meta of={withLongPressStories} />
+
+# withLongPress
+
+The `withLongPress` mixin can be wrapped around a parent component to handle key events and also to take advantage of the
+global `keyMetricsCallbackHandler` in the lightning UI Context.
```
ImCoolNowRight: Can we capitalize "lightning"?

ImCoolNowRight (commented):
Just a few thoughts on the mixin portion. I know the rest is still WIP.

packages/@lightningjs/ui-components/src/mixins/withLongPress/index.js
che-effe: Done

packages/@lightningjs/ui-components/src/mixins/withLongPress/index.js
che-effe: Done

packages/@lightningjs/ui-components/src/mixins/withLongPress/withLongPress.mdx
che-effe: removed

packages/@lightningjs/ui-components/src/mixins/withLongPress/index.js
```diff
@@ -0,0 +1,77 @@
...
+    static get name() {
+      return Base.name;
+    }
+    set threshold(value) {
+      this._threshold = value;
+    }
+
+    set executeOnce(val) {
+      this._executeOnce = val;
+    }
```
ImCoolNowRight: From our in-person discussion, let's change this to something like `continuousExecution` with a default value of `false` to make it clearer that the mixin allows for the signal to be fired repeatedly during a long press.

packages/@lightningjs/ui-components/src/mixins/withLongPress/withLongPress.stories.js
```diff
@@ -0,0 +1,89 @@
...
+  class withLongPress extends lng.Component {
+    static _template() {
+      return {
+        Tile: {
+          type: withLongPressMixin(Tile),
+          src: 'https://image.tmdb.org/t/p/w500/zHdQ6yaqDf3OQO5uhr0auAgwK6O.jpg',
+          w: 320,
+          h: 180,
+          threshold: args.threshold || 2000,
+          signals: { longPressHit: '_handleLongPress' }
```
ImCoolNowRight: We can remove this now that we're doing `fireAncestors` instead of a signal.

packages/@lightningjs/ui-components/src/mixins/withLongPress/index.js
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * this will handle only key up events
+     * */
+    _handleKeyRelease(keyEvent) {
+      this._pressedTimeStart = null;
+      this._hasExecuted = false;
+      super._handleKeyRelease(keyEvent);
+    }
+  };
```
ImCoolNowRight: I think we should check to see if there was a long press detected and call `fireAncestors` to let the parent know that the long press ended. Right now, the story has the text display for two seconds, but there's no way to link it so that it remains visible for the entirety of the long press.

packages/@lightningjs/ui-components/src/mixins/withLongPress/index.js
```diff
@@ -0,0 +1,77 @@
...
+      this._executeOnce = val;
+    }
+    _init() {
+      this._pressedTimeStart = null;
+      this._hasExecuted = false;
+      super._init();
+    }
+    /**
+     * this will handle only key down events
+     * it will grab a reference start time stamp and compare it to any subsuquent key down events' timestamp values
```
ImCoolNowRight: Should be "subsequent".

packages/@lightningjs/ui-components/src/mixins/withLongPress/index.js
```diff
@@ -0,0 +1,77 @@
...
+     * when we meet the set threshold time in seconds, we will execute a supplied callback method
+     * */
+    _handleKey(keyEvent) {
+      // capture the key event time stamp thr first time through to use as a reference.
```
ImCoolNowRight: "the"

packages/@lightningjs/ui-components/src/mixins/withLongPress/index.js
```diff
@@ -0,0 +1,77 @@
...
+        this._pressedTimeStart &&
+        keyEvent.timeStamp - this._pressedTimeStart >= (this._threshold || 2000)
+      ) {
+        // shortcircuit here if we only want to execute once before a key up event
+        if (this._executeOnce && this._hasExecuted) {
+          return;
+        }
+        this.fireAncestors('$longPressHit', keyEvent.key);
+        if (!this._executeOnce) {
+          this._pressedTimeStart = keyEvent.timeStamp;
```
ImCoolNowRight: I think it might make more sense to subtract the threshold instead so that we maintain any extra milliseconds between the threshold and when the key event occurs.

packages/@lightningjs/ui-components/src/mixins/withLongPress/index.js
ImCoolNowRight: Ok, ignore that second part about the story; I realized I didn't have `executeOnce` set to `false`. But I do still think `fireAncestors` would be useful to let the parent know that the press is done and behavior should change.

packages/@lightningjs/ui-components/src/mixins/withLongPress/withLongPress.stories.js
```diff
@@ -0,0 +1,89 @@
...
+        },
+        Notification: {
+          y: 220,
+          type: TextBox,
+          content: 'Long Press deteted',
```
ImCoolNowRight: "detected"

packages/@lightningjs/ui-components/src/mixins/withLongPress/withLongPress.stories.js
che-effe: Done!

ImCoolNowRight (approved):

ctoddy (approved):
l
g
t
m
😤

PR #533, opened 2024-09-11
feat(ScrollWrapper): add signals for button press at top and bottom of content

ImCoolNowRight:
## Description

Adds signal for Up/Down press when user is at top/bottom of ScrollWrapper container to allow for navigation if component is nested.

## References

LUI-1563

## Testing

Add the following code to the ScrollWrapper Basic story:

```js
// In the template's signals:
            onUpAtTop: '_handleUpAtTop',
            onDownAtBottom: '_handleDownAtBottom'

// As functions within the Basic class:
    _handleUpAtTop() {
      console.log('TOP');
    }

    _handleDownAtBottom() {
      console.log('BOTTOM');
    }
```

Open up the console and ensure that "TOP" and "BOTTOM" print when you are at the top and press Up or bottom and press Down.

## Automation

N/A

## Checklist

- [x] all commented code has been removed
- [x] any new console issues have been resolved
- [x] code linter and formatter has been run
- [x] test coverage meets repo requirements
- [x] PR name matches the expected semantic-commit syntax

PR #546, opened 2024-10-22
fix(ScrollWrapper): bubble focus events and remove signals

ImCoolNowRight:
## Description

The "fix" that was introduced in #533 was actually unnecessary. Instead, what was needed was bubbling the focus event up so that the parent component could handle it. This PR removes the signals and adds that bubbling.

(I'll remove the added story after getting approvals before merging this in.)

## References

LUI-1595

## Testing

Test the ScrollWrapper -> Focus story to make sure that you can navigate back up to the Buttons from the ScrollWrapper, but only once you have navigated back to the top of its content.

## Checklist


joshhowenstine (approved):
Makes sense to me

anthony9187 (approved):
L
G
T
M
🚢

PR #549, opened 2024-11-19
fix: createSharedReferences not comparing values

joshhowenstine:
## Description

There is a bug in which the first instance of a style property takes precedence over any additional occurrences of that property. For example, textStyle being used for one piece of text in a component will have its value used for every subsequent textStyle defined in the component's styles.

## References

[JIRA](https://ccp.sys.comcast.net/browse/LUI-1601)

## Checklist


packages/@lightningjs/ui-components/test.mjs
ImCoolNowRight, line 1: Can this file be removed?

ImCoolNowRight (approved):

ctoddy (approved):

